PR #9, opened 2020-11-16
Add azure pipelines agent vmss

simongottschlag:
* Added module for azure pipelines agent vmss


simongottschlag:
@phillebaba can you see if you can fix the failing check? My best guess is that the github action is using an option that just was deprecated.

simongottschlag:
https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

.github/workflows/check.yaml
```diff
@@ -7,7 +7,7 @@ jobs:
       - name: Clone repo
         uses: actions/checkout@master
       - name: Setup TFLint
-        uses: lablabs/setup-tflint@v1
+        uses: lablabs/setup-tflint@v1.0.2
```
simongottschlag, line 10: @phillebaba this fixes the issue.

PR #14, opened 2020-11-17
Feature/tfsec

### LINKED ISSUE #13: Evaluate tfsec
https://github.com/tfsec/tfsec

simongottschlag:
fixes #13


phillebaba:
I think we should have a tfsec recipe in the makefile so we can run it locally

simongottschlag:
> I think we should have a tfsec recipe in the makefile so we can run it locally

@phillebaba done!

phillebaba (approved):

PR #41, opened 2020-11-20
Update fluxcd-v1

simongottschlag:


phillebaba:
Why are we copying the whole fluxv1 chart?

PR #50, opened 2020-11-22
Feature/tekton operator

### LINKED ISSUE #49: Setup tekton
Setup terraform module for tekton.

The following will be required:
- [ ] Create module for `tekton pipelines` and `tekton triggers` (don't use `tekton-operator`, there seems to be issues when uninstalling/installing it)
- [ ] Create PR for tekton to add securityContext and volume mount to `initContainers` (or a mutating webhook using `opa`/`opa-gatekeeper`)

Putting issue on backlog.

simongottschlag:
fixes #49


modules/azure/aks/modules.tf
```diff
@@ -213,3 +213,13 @@ module "velero" {
   azure_client_id                 = var.velero.identity.client_id
   azure_resource_id               = var.velero.identity.resource_id
 }
+
+# Velero
+module "tekton-operator" {
```
NissesSenap, line 218: Does this have anything to do with velero?

validation/kubernetes/velero/main.tf
```diff
@@ -1,12 +1,20 @@
 terraform {
```
NissesSenap, line 1: Should this file be changed in this PR

NissesSenap (commented):
N00b questions :)

modules/azure/aks/modules.tf
simongottschlag: Fixed!

validation/kubernetes/velero/main.tf
simongottschlag: Not really, but fixed it since I saw that I've missed it. ;)

NissesSenap:
I know this is not what you are asking but in your task definition don't use: "inputs".
```
  steps:
  - name: product
    image: bash:latest
    script: |
      #!/usr/bin/env bash
      echo -n $(( "$(inputs.params.a)" * "$(inputs.params.b)" ))
```
It's not needed any more.
Just $(params.b) is enough

You also don't need the " around $(params.b) in your echo. And you have already defined that it's a string that you send in with your type.

simongottschlag:
Will abandon this PR and re-open at a later stage. We need to get tekton `initContainers` to work together with opa-gatekeeper policies and move away from tekton-operator to setup tekton pipelines and tekton triggers directly with terraform.

PR #58, opened 2020-11-26
Add support for AWS in external-dns

phillebaba:


modules/kubernetes/external-dns/charts/external-dns-extras/templates/azure-identity.yaml
```diff
@@ -1,3 +1,4 @@
+{{ if and .Values.resourceID .Values.clientID }}
```
simongottschlag, line 1: I think we should have some kind of standard of telling the modules what provider we are working with. Example: https://github.com/XenitAB/terraform-modules/blob/691da8d4cdd3916e394c0a9f8568f2405f0436cf/modules/kubernetes/velero/variables.tf#L1-L5

modules/kubernetes/external-dns/variables.tf
```diff
@@ -38,3 +38,15 @@ variable "azure_resource_id" {
...
+  description = "AWS region to use"
+  type        = string
+  default     = ""
+}
+
+variable "aws_role_arn" {
+  description = "AWS role ARN to attach to service account"
+  type        = string
+  default     = ""
+}
```
simongottschlag, line 52: I think it would be smart of us to have cloud-provider specific configuration in object, like:
```hcl
variable "aws_config" {
  description = "AWS role ARN to attach to service account"
  type        = object({aws_role_arn=string, aws_region=string})
  default     = {aws_role_arn="", aws_region=""}
}
```

Then we can have one object per provider and then it's always configured based on the cloud_provider string which one is used.

Thoughts?

modules/kubernetes/external-dns/templates/values.yaml.tpl
```diff
@@ -14,3 +14,10 @@ azure:
 podLabels:
   aadpodidbinding: external-dns
```
simongottschlag, line 15: This shouldn't be added when using aws.

modules/kubernetes/external-dns/main.tf
```diff
@@ -16,7 +16,15 @@ terraform {
 }
 
 locals {
-  values = templatefile("${path.module}/templates/values.yaml.tpl", { provider = var.dns_provider, sources = var.sources, azure_tenant_id = var.azure_tenant_id, azure_subscription_id = var.azure_subscription_id, azure_resource_group = var.azure_resource_group })
+  values = templatefile("${path.module}/templates/values.yaml.tpl", {
```
simongottschlag, line 19: What about loading different values-files based on the cloud provider?

simongottschlag (commented):
I've added a few comments.

modules/kubernetes/external-dns/charts/external-dns-extras/templates/azure-identity.yaml
phillebaba: I do not agree as external-dns works with more than just only cloud providers.

modules/kubernetes/external-dns/main.tf
phillebaba: I am fine with doing it this way for now, we can come back to this later if the values file becomes to crowded.

modules/kubernetes/external-dns/templates/values.yaml.tpl
phillebaba: It is not added when running aws?

modules/kubernetes/external-dns/variables.tf
phillebaba: done

simongottschlag (approved):
LGTM

PR #59, opened 2020-11-26
Add datadog

phillebaba:


modules/kubernetes/datadog/templates/values.yaml.tpl
```diff
@@ -0,0 +1,34 @@
+datadog:
+  apiKey: ${api_key}
+  clusterName: ${location}-${environment}
+  site: datadoghq.eu
```
simongottschlag, line 4: Make site a variable that defaults to datadoghq.eu

modules/kubernetes/datadog/templates/values.yaml.tpl
```diff
@@ -0,0 +1,34 @@
...
+    enabled: true
+    containerCollectAll: true
+    containerCollectUsingFiles: true
+  apm:
+    enabled: true
+  systemProbe:
+    enabled: false
+  env:
+    - name: DD_KUBELET_TLS_VERIFY
+      value: "false"
```
simongottschlag, line 23: Is this one still required?

simongottschlag (commented):
Two comments

modules/kubernetes/datadog/templates/values.yaml.tpl
phillebaba: no that should now be removed

modules/kubernetes/datadog/templates/values.yaml.tpl
phillebaba: sounds good

modules/kubernetes/datadog/templates/values.yaml.tpl
```diff
@@ -0,0 +1,31 @@
+datadog:
+  apiKey: ${api_key}
+  clusterName: ${location}-${environment}
+  site: ${datadog_site}datadoghq.eu
```
simongottschlag, line 4: This seems wrong

simongottschlag (approved):
LGTM

PR #60, opened 2020-11-26
Add AWS support to velero

phillebaba:


modules/kubernetes/velero/main.tf
```diff
@@ -20,53 +20,37 @@ terraform {
...
+  repository       = "https://vmware-tanzu.github.io/helm-charts"
+  chart            = "velero"
+  name             = "velero"
+  namespace        = "velero"
+  create_namespace = true
+  version          = "2.14.0"
+  values           = [local.values]
 }
 
 resource "helm_release" "velero_extras" {
```
simongottschlag, line 40: Should we really run this if we don't need it?

modules/kubernetes/velero/variables.tf
```diff
@@ -4,38 +4,37 @@ variable "cloud_provider" {
...
+    client_id                 = string,
+    resource_id               = string,
+    storage_account_name      = string,
+    storage_account_container = string
+  })
+  default = {
+    subscription_id           = "",
+    tenant_id                 = "",
+    resource_group            = "",
+    client_id                 = "",
```
simongottschlag, line 21: Why commas on only a few of them?

modules/kubernetes/velero/main.tf
phillebaba: we still need it as it adds the schedule

modules/kubernetes/velero/variables.tf
phillebaba: I will remove them

simongottschlag (approved):
LGTM

PR #62, opened 2020-11-27
Fix cert manager server url

phillebaba:


modules/kubernetes/cert-manager/variables.tf
```diff
@@ -3,3 +3,9 @@ variable "notification_email" {
   type        = string
   default     = "DG-Team-DevOps@xenit.se"
```
simongottschlag, line 4: Remove this default

modules/kubernetes/cert-manager/variables.tf
```diff
@@ -3,3 +3,9 @@ variable "notification_email" {
...
 }
+
+variable "server" {
```
simongottschlag, line 7: Is there a better name we can use?

simongottschlag (changes requested):

simongottschlag (approved):
LGTM

PR #96, opened 2020-12-18
Add graph

simongottschlag:


phillebaba:
Looks nice, even though a few of these modules creates really funny complex graphs :p

Issue #95, opened 2020-12-18
Migrate away from looping regions inside of the module


phillebaba:
We should consider moving away from looping regions inside of the modules before it is too late to make the switch. Doing the looping outside will greatly simplify the complexity as we dont have to generate env region variables.

phillebaba:
@simongottschlag

simongottschlag:
Finished:
- azure/governance: #124 
- azure/core: #125 
- azure/hub: #126

PR #111, opened 2021-02-05
Azure/bastion

tvuorenmaa89:
Added Azure Bastion module


modules/azure/bastion/locals.tf
```diff
@@ -0,0 +1,18 @@
+locals {
+  env_resources = { for region in var.regions : "${var.environment}-${region.location_short}-${var.name}" => {
```
phillebaba, line 2: We should move away from iterating regions in the module and instead to it outside the module.

@simongottschlag might be good to write some best practices regarding Azure Terraform?

simongottschlag: Agreed. We shouldn't build anything new that iterates regions in the modules since we want to move away from it: https://github.com/XenitAB/terraform-modules/issues/95

PR #121, opened 2021-02-15
Add secret store csi unique volume constraint

phillebaba:
Adds a new default constraint that will be applied when deploying OPA. Will have no affect on those who do not require CSI and does not require custom apps or CRDs to be deployed for it to work.


simongottschlag (approved):
LGTM

PR #126, opened 2021-02-22
azure/hub: Remove region loop

simongottschlag:
Changes:
* Remove region loop from azure/hub


phillebaba (approved):

simongottschlag:
Changes:
- Remove region loop from azure/hub

PR #128, opened 2021-02-23
Make version values explicit

### LINKED ISSUE #122: Make Helm versions explicit
Now when auto updating providers and Helm charts in the modules, we need to have the versions of the Helm charts explicit and not as locals or variables.

phillebaba:
Moves version values from locals to be set in the `helm_release` resource.

Fixes #122


simongottschlag (approved):
LGTM

PR #132, opened 2021-02-24
Disable priority class

phillebaba:
This disables the priority class set by default in the gatekeeper Helm chart which is `system-node-critical`. The reason this needs to be done is because only "newer" >1.17 clusters will allow use of this priority class outside of the kube-system namespace. This change should be reverted when all clusters run the latest version of Kubernetes.

https://github.com/kubernetes/kubernetes/issues/78383


simongottschlag (commented):
LGTM

PR #186, opened 2021-03-25
Add prometheus module

NissesSenap:
The module will install the prometheus-operator + a number of scrapers
cluster roles etc.
According to https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack


modules/kubernetes/prometheus/main.tf
```diff
@@ -0,0 +1,43 @@
...
+resource "kubernetes_namespace" "this" {
+  metadata {
+    labels = {
+      name = local.namespace
+    }
+    name = local.namespace
+  }
+}
+
+resource "helm_release" "datadog" {
```
phillebaba, line 36: Change to prometheus

modules/kubernetes/aks-core/modules.tf
```diff
@@ -308,3 +308,24 @@ module "azad_kube_proxy" {
     scope         = var.azad_kube_proxy_config.k8dash_config.scope
   }
 }
+
+# Prometheus
+module "prometheus" {
+  depends_on = [kubernetes_namespace.tenant, module.opa_gatekeeper]
```
simongottschlag, line 314: Why does it depend on the tenant namespaces?

modules/kubernetes/aks-core/modules.tf
```diff
@@ -308,3 +308,24 @@ module "azad_kube_proxy" {
...
+    if var.prometheus_enabled
+  }
+
+  source = "../../kubernetes/prometheus"
+
+  remote_write_enabled = var.prometheus_config.remote_write_enabled
+  remote_write_url     = var.prometheus_config.remote_write_url
+  remote_write_name    = var.prometheus_config.remote_write_name
+
+  volume_claim_enabled            = var.prometheus_config.volume_claim_enabled
```
simongottschlag, line 328: Check up on how AKS handles disks and AZs. Will Prometheus be bound to a specific AZ when using a PV? Do we need to use a specific storage?

modules/kubernetes/prometheus/charts/prometheus-extras/values.yaml
```diff
@@ -0,0 +1,24 @@
...
+resources: {}
+
+remoteWrite:
+  enabled: true
+  url: ""
+  name: xenitInfra
+
+volumeClaim:
+  enabled: true
+  storageClassName: default
```
simongottschlag, line 23: As mentioned above, check up on how this will affect things and if it will be bound to a specific AZ

modules/kubernetes/prometheus/templates/values.yaml.tpl
```diff
@@ -0,0 +1,20 @@
...
+  enabled: false
+
+# We don't use alert manager in the cluster, we use thanos ruler
+alertmanager:
+  enabled: false
+
+prometheus:
+  enabled: false
+
+kube-state-metrics:
```
simongottschlag, line 18: Verify that AKS doesn't setup one

modules/kubernetes/prometheus/templates/values-extras.yaml.tpl
```diff
@@ -0,0 +1,13 @@
+# Default values for prometheus-extras.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+remoteWrite:
```
simongottschlag, line 5: Config regarding client cert

modules/kubernetes/prometheus/main.tf
```diff
@@ -0,0 +1,59 @@
+/**
```
simongottschlag, line 1: Maybe we should read the client cert and key from KeyVault inside of this module?

simongottschlag (commented):
A few comments. 😊👍

modules/kubernetes/aks-core/modules.tf
NissesSenap: When you write specific AZ do you mean a specific azure zone?

Specifically about zones: https://docs.microsoft.com/en-us/azure/aks/availability-zones#azure-disks-limitations
```
Volumes that use Azure managed disks are currently not zone-redundant resources. Volumes cannot be attached across zones and must be co-located in the same zone as a given node hosting the target pod.

Kubernetes is aware of Azure availability zones since version 1.12. You can deploy a PersistentVolumeClaim object referencing an Azure Managed Disk in a multi-zone AKS cluster and Kubernetes will take care of scheduling any pod that claims this PVC in the correct availability zone.
```

So my understanding is that node 1 is scheduled in zone A and node 2 is scheduled in zone B the disks will be created in those zones. We just have to write a anti affinity rule that says that we shoulden't shedule prometheus in the same zones.
We should be able to use label: topology.kubernetes.io/zone or failure-domain.beta.kubernetes.io/zone.
Here I need someones AKS expertiese to tell me witch one is the best.

PVC info: https://docs.microsoft.com/en-us/azure/aks/concepts-storage
```
managed-csi - Uses Azure StandardSSD locally redundant storage (LRS) to create a Managed Disk. The reclaim policy ensures that the underlying Azure Disk is deleted when the persistent volume that used it is deleted. The storage class also configures the persistent volumes to be expandable, you just need to edit the persistent volume claim with the new size.
```

And from: https://docs.microsoft.com/en-us/azure/aks/azure-disks-dynamic-pv
```
Each AKS cluster includes four pre-created storage classes, two of them configured to work with Azure disks:

The default storage class provisions a standard SSD Azure disk.
Standard storage is backed by Standard SSDs and delivers cost-effective storage while still delivering reliable performance.
The managed-premium storage class provisions a premium Azure disk.
Premium disks are backed by SSD-based high-performance, low-latency disk. Perfect for VMs running production workload. If the AKS nodes in your cluster use premium storage, select the managed-premium class.

```

This disk is only for temporary storage, it will send all the data to Thanos. But if our pods restart we shoulden't lose the data like we would with a emptyDir.
But we also don't need to think about backups or anything like that, if we losse the data we should max loose an hour of data.

If we get higher demands from our customers we have the possability to use other storageClasses.

modules/kubernetes/prometheus/templates/values.yaml.tpl
NissesSenap: Setup what?

modules/kubernetes/aks-core/modules.tf
NissesSenap: Found docs about failure-domain.beta.kubernetes.io/zone and it's deprecated: https://kubernetes.io/docs/reference/labels-annotations-taints/#failure-domainbetakubernetesiozone

We should use topology.kubernetes.io/zone

modules/kubernetes/prometheus/charts/prometheus-extras/templates/prometheus.yaml
```diff
@@ -0,0 +1,43 @@
...
+      preferredDuringSchedulingIgnoredDuringExecution:
+        - weight: 100
+          podAffinityTerm:
+            labelSelector:
+              matchExpressions:
+                - key: prometheus
+                  operator: In
+                  values:
+                    - prometheus-xks
+            topologyKey: topology.kubernetes.io/zone
```
NissesSenap, line 39: @simongottschlag see anyi-affinity rule

modules/kubernetes/aks-core/modules.tf
NissesSenap: Created anti-affinity rule, see bellow in tagged comment.

modules/kubernetes/prometheus/charts/prometheus-extras/values.yaml
NissesSenap: Created anti-affinity rule, see bellow in tagged comment + managment of aks disk managment

modules/kubernetes/aks-core/modules.tf
```diff
@@ -134,6 +134,8 @@ module "ingress_nginx" {
   }
 
   source = "../../kubernetes/ingress-nginx"
+
+  http_snipet = var.ingress_config.http_snipet
```
NissesSenap, line 138: I might have gotten one to many commits in to this PR....

modules/kubernetes/prometheus/templates/values.yaml.tpl
simongottschlag: kube-state-metrics

modules/kubernetes/aks-core/modules.tf
```diff
@@ -310,3 +310,27 @@ module "azad_kube_proxy" {
...
+  }
+
+  source = "../../kubernetes/prometheus"
+
+  alertmanager_enabled = var.prometheus_config.alertmanager_enabled
+
+  remote_write_enabled   = var.prometheus_config.remote_write_enabled
+  remote_write_url       = var.prometheus_config.remote_write_url
+  remote_write_name      = var.prometheus_config.remote_write_name
+  remote_tls_secret_name = var.prometheus_config.remote_tls_secret_name #tfsec:ignore:GEN003
```
simongottschlag, line 331: Add comment that we're ignoring since it's `secret_name` and not the `secret`.

modules/kubernetes/prometheus/charts/prometheus-extras/templates/alertmanager.yaml
```diff
@@ -0,0 +1,46 @@
...
+    groupInterval: 5m
+    repeatInterval: 12h
+    receiver: 'opsgenie'
+  receivers:
+  - name: 'opsgenie'
+    opsgenieConfigs:
+    - apiURL: 'https://api.eu.opsgenie.com/'
+      apiKey:
+        name: 'opsgenie-config'
+        key: 'apiSecret'
```
simongottschlag, line 45: We should make opsgenie configurable using variables. `apiURL` should be a variable in Helm that we inject from terraform. apiKey.key and apiKey.name should alsoe be variables.

modules/kubernetes/prometheus/charts/prometheus-extras/templates/rules.yaml
```diff
@@ -0,0 +1,15 @@
...
+    prometheus: prometheus-xks
+    role: alert-rules
+    platform: "true"
+  name: prometheus-example-rules
+spec:
+  groups:
+  - name: ./example.rules
+    rules:
+    - alert: ExampleAlert
+      expr: vector(1)
```
simongottschlag, line 15: example?

modules/kubernetes/prometheus/charts/prometheus-extras/values.yaml
```diff
@@ -0,0 +1,27 @@
...
+service:
+  type: ClusterIP
+  port: 9090
+
+resources: {}
+
+remoteWrite:
+  enabled: true
+  url: ""
+  name: xenitInfra
```
simongottschlag, line 19: should this really be the default? maybe use something more generic than xenit?

modules/kubernetes/prometheus/variables.tf
```diff
@@ -0,0 +1,46 @@
...
+  type        = string
+}
+
+variable "remote_write_name" {
+  description = "The name of the remote write in prometheus"
+  type        = string
+  default     = "xenitInfra"
+}
+
+variable "remote_tls_secret_name" {
```
simongottschlag, line 24: Where do we configure this secret? Maybe we should include csi secret to be able to read it from KeyVault?

simongottschlag (commented):
A few comments

modules/kubernetes/prometheus/variables.tf
NissesSenap: Currently we assume that it's there. I started to look in to csi secret to manage it.

modules/kubernetes/prometheus/charts/prometheus-extras/values.yaml
NissesSenap: Any suggestions? 
Just infra, thanos?

modules/kubernetes/prometheus/charts/prometheus-extras/templates/rules.yaml
NissesSenap: Should be removed, thanks

modules/kubernetes/prometheus/charts/prometheus-extras/values.yaml
simongottschlag: +1 for `thanos`

modules/kubernetes/prometheus/charts/prometheus-extras/templates/prometheus.yaml
```diff
@@ -0,0 +1,73 @@
...
+  namespace: prometheus
+spec:
+  replicas: {{ .Values.replicaCount }}
+  serviceAccountName: {{ .Values.serviceAccount.name }}
+  serviceMonitorSelector:
+    matchLabels:
+      platform: "true"
+  probeSelector:
+    matchLabels:
+      platform: "true"
```
phillebaba, line 16: Change to something more unique

modules/kubernetes/prometheus/charts/prometheus-extras/templates/prometheus.yaml
```diff
@@ -0,0 +1,73 @@
...
+  labels:
+    {{- include "prometheus-extras.labels" . | nindent 4 }}
+  name: prometheus-xks
+  namespace: prometheus
+spec:
+  replicas: {{ .Values.replicaCount }}
+  serviceAccountName: {{ .Values.serviceAccount.name }}
+  serviceMonitorSelector:
+    matchLabels:
+      platform: "true"
```
phillebaba, line 13: Change to something more unique

phillebaba (approved):

simongottschlag (approved):
LGTM

PR #199, opened 2021-04-08
Add ingress healthz

phillebaba:
Adds ingress healthz as a module to aks-core.


modules/kubernetes/ingress-healthz/templates/values.yaml.tpl
```diff
@@ -0,0 +1,13 @@
+application:
+  environment: ${environment}
+
```
phillebaba, line 3: need a good way of setting the domain and certificate issuer.

simongottschlag (changes requested):
We need to add ingress to the helm chart before merging this

NissesSenap:
Can you explain what you mean @simongottschlag "We need to add ingress to the helm chart before merging this"
There is a ingress template in the helm chart.
https://github.com/XenitAB/ingress-healthz/blob/master/charts/ingress-healthz/templates/ingress.yaml

modules/kubernetes/ingress-healthz/templates/values.yaml.tpl
NissesSenap: Yeah other then that I think it looks good.

simongottschlag:
> Can you explain what you mean @simongottschlag "We need to add ingress to the helm chart before merging this"
> There is a ingress template in the helm chart.
> https://github.com/XenitAB/ingress-healthz/blob/master/charts/ingress-healthz/templates/ingress.yaml

never mind :-)

simongottschlag (approved):
LGTM

PR #203, opened 2021-04-09
Fix breaking values change in aad-pod-id chart

phillebaba:
Fixes breaking change in chart values.
https://azure.github.io/aad-pod-identity/docs/#v17x-breaking-change


NissesSenap (approved):

NissesSenap:
lgtm

PR #213, opened 2021-04-11
Add monitor resources

phillebaba:
Adds `PodMonitor`and `ServiceMonitor`resources to most of the kubernetes modules. There may be some module that has been missed but this is a good starting point to get things going.

Additionally all namespaces have received the label `xkf.xenit.io/kind` with the value of `platform` or `tenant` to simplify creating selectors for specific namespaces.

### DIFF
--- a/modules/kubernetes/opa-gatekeeper/main.tf
+++ b/modules/kubernetes/opa-gatekeeper/main.tf
@@ -112,7 +112,7 @@ resource "kubernetes_namespace" "this" {
     labels = {
       name                             = "gatekeeper-system"
       "admission.gatekeeper.sh/ignore" = "no-self-managing"
-      "xkf.xenit.io/kind" = "platform"
+      "xkf.xenit.io/kind"              = "platform"
     }
     name = "gatekeeper-system"
   }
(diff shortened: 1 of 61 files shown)


modules/kubernetes/opa-gatekeeper/main.tf
```diff
@@ -133,7 +134,7 @@ resource "helm_release" "gatekeeper_templates" {
   chart      = "gatekeeper-library-templates"
   name       = "gatekeeper-library-templates"
   namespace  = kubernetes_namespace.this.metadata[0].name
-  version    = "v0.7.0"
+  version    = "v0.7.1"
```
NissesSenap, line 137: Note the helm chart update. Not a problem but good to know about

NissesSenap (commented):
It looks great but my question is why podMonitor instead of serviceMonitor?
In my experience the default is serviceMonitor and podMonitor is something that you might use on strage stateful sets like kafka.

simongottschlag (approved):
LGTM

modules/kubernetes/opa-gatekeeper/main.tf
phillebaba: Needed as I found a bug in the exclude namespace logic.

phillebaba:
@NissesSenap PodMonitors are used when services do not exist. It feels weird to create a service just for the metrics collection. If I am wrong we can change to use ServiceMonitor later.

NissesSenap:
> @NissesSenap PodMonitors are used when services do not exist. It feels weird to create a service just for the metrics collection. If I am wrong we can change to use ServiceMonitor later.

Yeah this I'm aware about, just assumed that most of those pods had a service .
But took a quick look now and apparently not :) 

Assumptions is the base of all fuck ups

PR #222, opened 2021-04-14
Add resource limits on prometheus

NissesSenap:
These values is currently hard-coded.
It's supposed to only look at XKF features.

### DIFF
--- a/modules/kubernetes/prometheus/templates/values-extras.yaml.tpl
+++ b/modules/kubernetes/prometheus/templates/values-extras.yaml.tpl
@@ -18,4 +18,3 @@ resources:
     cpu: "20m"
   limits:
     memory: "500Mi"
-    cpu: "100m"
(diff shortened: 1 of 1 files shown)


modules/kubernetes/prometheus/templates/values-extras.yaml.tpl
```diff
@@ -11,3 +11,11 @@ volumeClaim:
...
   storageClassName: ${volume_claim_storage_class_name}
   size: ${volume_claim_size}
+
+resources:
+  requests:
+    memory: "250Mi"
+    cpu: "20m"
+  limits:
+    memory: "500Mi"
+    cpu: "100m"
```
phillebaba, line 21: Lets not set cpu limits as it only artificially throttles CPU.

NissesSenap: Sorry old habits die hard

phillebaba (approved):

PR #223, opened 2021-04-14
Set container security context for ingress-healthz

phillebaba:
Removes the need to rely on mutation webhooks for setting a correct security context.


NissesSenap (approved):
LGTM

PR #225, opened 2021-04-15
Add ingress nginx name override and internal setting

phillebaba:
Adds options to set name override, which is appended to the end of the name, and internal lb settings.

This is to enable public/private ingress use cases. The name override will be used in the namespace, deployment name, and ingress class. If nothing is set the behavior will be the same as before.


NissesSenap (approved):
LGTM

PR #241, opened 2021-04-23
Update prometheus external labels

phillebaba:


NissesSenap (approved):
If we hit labeling issues with environment I'm blaming you :D

PR #244, opened 2021-04-26
Update metrics collection

phillebaba:
* Add option to explicitly set tenant_id
* Expose kind value for monitor selectors
* Enable metrics collection on nginx ingress
* Remove unique env-cluster external label


modules/kubernetes/prometheus/templates/values-extras.yaml.tpl
```diff
@@ -5,6 +5,8 @@
 remoteWrite:
   enabled: ${remote_write_enabled}
   url: ${remote_write_url}
+  headers:
+    THANOS-TENANT: ${tenant_id}
```
simongottschlag, line 9: Make optional, only add if tenant_id is set

simongottschlag (dismissed):

NissesSenap (approved):
LGTM

PR #247, opened 2021-04-27
Change falco to use metrics exporter instead of sidekick

phillebaba:
Changes falco to export events with metrics.


simongottschlag (approved):
LGTM

PR #251, opened 2021-04-29
Remove version pinning from flux

phillebaba:
The flux provider pins the flux version by default now, so there is no need to do this in the module also.


NissesSenap (approved):
LGTM

PR #263, opened 2021-05-07
Typo

madakralc:
Found a typo whilst reading file


simongottschlag:
@madakralc you need to change the following line: https://github.com/XenitAB/terraform-modules/blob/8a0b60c49af83913856c100e2f84c3d0b23aabbc/modules/azure/hub/variables.tf#L58

Then run `make docs` and commit.

Thank you :+1:

PR #269, opened 2021-05-12
Add goldpinger

phillebaba:
This change adds goldpinger with a servicemonitor to scrape its metrics.

Closes #269

### DIFF
--- a/modules/kubernetes/aks-core/modules.tf
+++ b/modules/kubernetes/aks-core/modules.tf
@@ -383,5 +383,7 @@ module "goldpinger" {
   }
 
   source = "../../kubernetes/goldpinger"
+
+  linkerd_enabled = var.linkerd_enabled
 }
 
--- a/modules/kubernetes/aks-core/modules.tf
+++ b/modules/kubernetes/aks-core/modules.tf
@@ -347,7 +347,7 @@ module "prometheus" {
   resource_selector  = var.prometheus_config.resource_selector
   namespace_selector = var.prometheus_config.namespace_selector
 
-  linkerd_enabled = var.linkerd_enabled
+  linkerd_enabled    = var.linkerd_enabled
   goldpinger_enabled = var.goldpinger_enabled
 }
 
--- a/modules/kubernetes/aks-core/modules.tf
+++ b/modules/kubernetes/aks-core/modules.tf
@@ -348,6 +348,7 @@ module "prometheus" {
   namespace_selector = var.prometheus_config.namespace_selector
 
   linkerd_enabled = var.linkerd_enabled
+  goldpinger_enabled = var.goldpinger_enabled
 }
 
 # xenit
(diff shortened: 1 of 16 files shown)


modules/kubernetes/aks-core/modules.tf
```diff
@@ -372,3 +373,15 @@ module "xenit" {
   loki_api_fqdn        = var.xenit_config.loki_api
 }
 
+module "goldpinger" {
+  depends_on = [module.opa_gatekeeper, module.linkerd]
```
NissesSenap, line 377: Do we need linkerd to run goldpinger? Can't we use this on all our clusters for network monitoring?

modules/kubernetes/aks-core/modules.tf
```diff
@@ -372,3 +373,15 @@ module "xenit" {
   loki_api_fqdn        = var.xenit_config.loki_api
 }
 
+module "goldpinger" {
```
simongottschlag, line 376: Make linkerd optional

modules/kubernetes/goldpinger/templates/values.yaml.tpl
```diff
@@ -0,0 +1,7 @@
+goldpinger:
+  port: 8080
+
+podAnnotations:
+  linkerd.io/inject: enabled
```
simongottschlag, line 8: make this optional and based on the linkerd_enabled var

modules/kubernetes/prometheus/charts/prometheus-extras/values.yaml
```diff
@@ -41,3 +41,4 @@ prometheus:
 
 enabledMonitors:
   linkerd: false
+  goldpinger: false
```
simongottschlag, line 44: enabled by default?

modules/kubernetes/goldpinger/templates/values.yaml.tpl
NissesSenap, line 8: Add a if statment check if linkerd enabled or not

modules/kubernetes/aks-core/modules.tf
NissesSenap: Is terrafrom smart enough that if linkerd module is disabled it will still continue?

NissesSenap (commented):
Minor comments

modules/kubernetes/aks-core/modules.tf
phillebaba: Yes terraform will be smart enough to ignore this if linkerd is not enabled.

modules/kubernetes/goldpinger/templates/values.yaml.tpl
phillebaba: Done

phillebaba: Done

modules/kubernetes/aks-core/modules.tf
phillebaba: Done

simongottschlag (approved):
Great work @phillebaba ! :100:

PR #275, opened 2021-05-17
Change alertmanager ns monitor

NissesSenap:


modules/kubernetes/prometheus/templates/values-extras.yaml.tpl
```diff
@@ -40,6 +40,19 @@ prometheus:
...
 alertmanager:
   enabled: ${alertmanager_enabled}
+    ruleSelector:
+    matchExpressions:
+      - key: xkf.xenit.io/monitoring
+        operator: In
+        values: ${resource_selector}
+      - key: xkf.xenit.io/rule
+        operator: In
+        values: prometheus
```
NissesSenap, line 50: Do we want to hardcode this or should we use it as a variable?

phillebaba (approved):

PR #278, opened 2021-05-18
Set priority class name

phillebaba:
This change will add some priority classes to the cluseter in aks-core and then also use these in the different modules. Some helm charts do currently not support setting priority class, I will create PRs to add the value but we should merge this instead of waiting.

Missing charts:
* azdo-proxy
* gatekeeper (Fixed in open-policy-agent/gatekeeper/pull/1266)
* reloader
* ingress-healthz (bitnami nginx)


modules/kubernetes/aks-core/k8s-priority-class.tf
```diff
@@ -0,0 +1,88 @@
+# Priority classes allows pods to be scheduled before other pods and evict pods from nodes.
+# There are two types of priority classes, platform and tenant. All platform priority classes
+# should have a higher priority value than the tenant priority classes. The platform-high
+# class is not set to the maximum value on purpose, as to leave space to create even more
+# prioritized classes in the future. It is preferred to use these classes rather than
+# system-cluster-critical and system-node-critical as they are used by AKS critical pods
+# which should have higher priority.
+#
+# Hot tip if you want to list all pods and their priority class in a cluster.
```
NissesSenap, line 9: Love the Hot tip

NissesSenap (approved):
We are also missing:
- linkerd
- datadog
- fluxcd v1
- velero

I still think we can merge this one create a seperate PR.
Great job

PR #292, opened 2021-05-27
Add allowed priority names

phillebaba:
Temp fix for #293.


NissesSenap (approved):
Temporary solution. It's okay

PR #298, opened 2021-06-04
Enable tenant users to use more CR:s

NissesSenap:
* Flux
* Cert-manager


bittrance:
It looks like this is two different changes in the same commit? If the labels are unrelated, it would be good if the PR explained why they are set.

modules/kubernetes/aks-core/k8s-cluser-role-binding-list-ns.tf
```diff
@@ -29,7 +30,8 @@ resource "kubernetes_cluster_role_binding" "edit_list_ns" {
     name = "${each.value.name}-edit-listns"
 
     labels = {
-      "aadGroup" = var.aad_groups.edit[each.key].name
+      "aadGroup"          = var.aad_groups.edit[each.key].name
```
phillebaba, line 33: Is there a reason this is different to the others?

modules/kubernetes/aks-core/k8s-cluster-role.tf
```diff
@@ -20,24 +26,30 @@ resource "kubernetes_cluster_role" "helm_release" {
...
   metadata {
-    name = "toolkit-helm-release"
+    name = "toolkit-edit"
+    labels = {
+      "xkf.xenit.io/kind" = "platform"
+    }
   }
   rule {
-    api_groups = ["helm.toolkit.fluxcd.io"]
+    api_groups = ["helm.toolkit.fluxcd.io", "kustomize.toolkit.fluxcd.io", "source.toolkit.fluxcd.io", "notification.toolkit.fluxcd.io"]
```
bittrance, line 37: Multi-line formatting would be easier to read, i think.

NissesSenap:
I can only agree with you @bittrance sadly we solve lots of things by talking and it makes us lazy when creating a PR:s from time to time. I have updated the PR descritpion to make it a bit clearer.

modules/kubernetes/aks-core/k8s-cluster-role.tf
```diff
@@ -20,24 +26,30 @@ resource "kubernetes_cluster_role" "helm_release" {
...
+    labels = {
+      "xkf.xenit.io/kind" = "platform"
+    }
   }
   rule {
-    api_groups = ["kustomize.toolkit.fluxcd.io"]
+    api_groups = ["datadogmetrics.datadoghq.com", "datadogmonitors.datadoghq.com"]
     resources  = ["*"]
-    verbs      = ["get, list, watch"]
+    verbs      = ["*"]
```
phillebaba, line 53: I thought we agreed to add all rules in the same cluster role? That way we dont need a role binding per rule.

modules/kubernetes/aks-core/k8s-cluster-role.tf
```diff
@@ -20,24 +26,30 @@ resource "kubernetes_cluster_role" "helm_release" {
   }
 }
 
-resource "kubernetes_cluster_role" "toolkit_helm_release" {
+resource "kubernetes_cluster_role" "toolkit_edit" {
   metadata {
-    name = "toolkit-helm-release"
+    name = "toolkit-edit"
```
bittrance, line 31: "Toolkit" is a fairly common word. Given that this is global naming, is it unique enough?

modules/kubernetes/aks-core/k8s-cluser-role-binding-list-ns.tf
NissesSenap: What's different? These rules are cluster-role-binding and not role-bindings.

modules/kubernetes/aks-core/k8s-cluster-role.tf
NissesSenap: I put all of them together and updated the name to custom_cr_edit

modules/kubernetes/aks-core/k8s-cluster-role.tf
NissesSenap: Done

modules/kubernetes/aks-core/k8s-cluster-role.tf
NissesSenap: Done

modules/kubernetes/aks-core/k8s-cluser-role-binding-list-ns.tf
NissesSenap: Nice catch, not to my knowladge. I updated it to match now.

phillebaba (approved):

PR #300, opened 2021-06-06
Update azdo proxy to latest version

phillebaba:
This change updates azdo proxy to the latest version. The major change is that token generation is no longer done by Terraform but rather in the app.

### DIFF
--- a/modules/kubernetes/fluxcd-v2-azdo/main.tf
+++ b/modules/kubernetes/fluxcd-v2-azdo/main.tf
@@ -75,7 +75,7 @@ resource "helm_release" "azdo_proxy" {
   chart      = "azdo-proxy"
   name       = "azdo-proxy"
   namespace  = kubernetes_namespace.this.metadata[0].name
-  version    = "v0.4.0-rc1"
+  version    = "v0.4.0"
   values = [templatefile("${path.module}/templates/azdo-proxy-values.yaml.tpl", {
     azure_devops_pat  = var.azure_devops_pat,
     azure_devops_org  = var.azure_devops_org,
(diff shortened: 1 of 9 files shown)


modules/kubernetes/fluxcd-v2-azdo/templates/azdo-proxy-values.yaml.tpl
```diff
@@ -1,6 +1,5 @@
-image:
-  tag: v0.3.6
```
NissesSenap, line 2: Why not define image.tag in the values file any more?

modules/kubernetes/fluxcd-v2-azdo/main.tf
```diff
@@ -99,11 +97,6 @@ data "azuredevops_git_repository" "cluster" {
   name       = var.cluster_repo
 }
 
-resource "random_password" "cluster" {
-  length  = 32
```
NissesSenap, line 103: random_password is set by the application now a days if i remember correctly

modules/kubernetes/fluxcd-v2-azdo/templates/azdo-proxy-values.yaml.tpl
phillebaba: This was wrong from the beginning and should be picked up by the application/chart version instead.

modules/kubernetes/fluxcd-v2-azdo/main.tf
phillebaba: Yes, all password generation has been removed from the Terraform now.

simongottschlag (approved):
LGTM :)

PR #301, opened 2021-06-07
Fix priority class names

phillebaba:
This change adds or fixes priority class names to the following modules:
* xenit
* ingress-healthz
* reloader
* azdo-proxy


NissesSenap (approved):
LGTM

PR #304, opened 2021-06-08
Ignore azurerm update

phillebaba:
There is currently a bug in azurerm version 2.62.0.
https://github.com/terraform-providers/terraform-provider-azurerm/issues/12060

We need to ignore updates of the provider until it is fixed and released. Generally new azurerm versions are released on fridays so this should be kept at least until then.

### DIFF
--- a/modules/azure/aks-global/main.tf
+++ b/modules/azure/aks-global/main.tf
@@ -8,6 +8,7 @@ terraform {
   required_version = "0.15.3"
 
   required_providers {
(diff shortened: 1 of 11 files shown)


NissesSenap (approved):
LGTM

simongottschlag:
@phillebaba @NissesSenap https://github.com/terraform-providers/terraform-provider-azurerm/releases/tag/v2.62.1

PR #329, opened 2021-07-05
Refactor namespaces into single file

phillebaba:
This changes does a quick refactor and gathers all namespace resources into a single file. Additionally it updates the network policy to allow DNS traffic.


NissesSenap (approved):
LGTM

PR #332, opened 2021-07-08
Add kube2iam helm chart

NissesSenap:


modules/kubernetes/kube2iam/templates/values.yaml.tpl
```diff
@@ -0,0 +1,38 @@
...
+  interface: cali+
+  port: 8181
+
+prometheus:
+  # Port to expose the /metrics endpoint on. If unset, defaults to `host.port`
+  # metricsPort: 9543
+  service:
+    enabled: true
+
+  serviceMonitor:
```
phillebaba, line 23: Remove and define in prometheus module

modules/kubernetes/kube2iam/templates/values.yaml.tpl
```diff
@@ -0,0 +1,38 @@
+extraArgs:
+  namespace-restrictions: true
+
+#   base-role-arn: arn:aws:iam::0123456789:role/
```
phillebaba, line 4: no need to have any of these

modules/kubernetes/kube2iam/templates/values.yaml.tpl
```diff
@@ -0,0 +1,38 @@
...
+    # Labels to add to the service monitor
+    labels:
+      xkf.xenit.io/monitoring: platform
+
+priorityClassName: "platform-high"
+
+rbac:
+  ## If true, create & use RBAC resources
+  ##
+  create: true
```
phillebaba, line 35: is this the same as the default value? If so it can be removed

phillebaba (commented):
Some small comments, but you need to add the module to eks-core.

modules/kubernetes/kube2iam/templates/values.yaml.tpl
NissesSenap: done

modules/kubernetes/kube2iam/templates/values.yaml.tpl
NissesSenap: psp is deafult but not rbac.

NissesSenap:
We have decided to not go with kube2iam

PR #334, opened 2021-07-08
Enable external-dns, cert-manager and ingress-nginx for AWS

NissesSenap:
This change fixes issues when deploying external-dns, cert-manager and ingress-nginx in an XKF EKS cluster. Most of the issues are related to Kubernetes webhooks not working without host networking when running EKS with Calico.


modules/aws/eks/iam.tf
```diff
@@ -0,0 +1,125 @@
...
+    ]
+    resources = ["*"]
+  }
+}
+
+module "external_dns" {
+  source = "../irsa"
+
+  name = "external-dns-${var.name}${var.eks_name_suffix}"
+  oidc_providers = [
```
simongottschlag, line 68: Not a requirement, but wouldn't it be nice to make this an array of maps? In that way, we would be able to see that it belongs to eks or eks1 or eks2 as an example.

modules/aws/irsa/variables.tf
```diff
@@ -3,9 +3,12 @@ variable "name" {
   type        = string
 }
 
-variable "oidc_urls" {
-  description = "EKS OIDC URLs to trust"
-  type        = list(string)
+variable "oidc_providers" {
```
simongottschlag, line 6: I wrote this above, but could it be logical to make it a list of maps? To easier see which cluster it belongs to (as an example).

phillebaba (approved):

PR #336, opened 2021-07-09
Add hostNetwork config option to ingress-nginx

NissesSenap:
This to be able to support running on EKS with calico CNI


NissesSenap:
This is handled in https://github.com/XenitAB/terraform-modules/pull/334

PR #342, opened 2021-07-10
Introduce deny default networkpolicy to all namespaces

phillebaba:
This change will add a network policy to all platform namespaces and then additional polices to allow traffic that is required.

There are currently two limitations with this solution which is for the most part caused by running managed clusters.
* Access to the api server cant be limited to just the specific IP as it is dynamic and resolved through DNS. OSS Calico does not support DNS based egress rules and using the IP from Terraform would be risky if that IP changes in the middle of the night. The current solution is to allow pods that need api server access egress permission to all public IPs. We shoudl re visit this in the future.
* Controller which implement Kubernetes webhooks have to allow traffic from the api server. I think I have figured out the CIDR used in AKS but we should keep an eye on this to make sure that this is correct all of the time. We still need to figure out the network policy to use in EKS to make this possible,


modules/kubernetes/ingress-healthz/main.tf
```diff
@@ -31,6 +31,88 @@ resource "kubernetes_namespace" "this" {
...
+        pod_selector {
+          match_labels = {
+            k8s-app = "kube-dns"
+          }
+        }
+      }
+
+      ports {
+        port     = 53
+        protocol = "UDP"
```
simongottschlag, line 65: @phillebaba we should allow both UDP and TCP: https://datatracker.ietf.org/doc/html/rfc6891#section-4.3

modules/kubernetes/ingress-nginx/main.tf
```diff
@@ -29,6 +29,133 @@ resource "kubernetes_namespace" "this" {
...
+
+    ingress {
+      from {
+        ip_block {
+          cidr = "0.0.0.0/0"
+          except = [
+            "10.0.0.0/8",
+            "172.16.0.0/12",
+            "192.168.0.0/16",
+          ]
```
simongottschlag, line 140: Could this cause issues for internal ingress?

modules/kubernetes/ingress-nginx/main.tf
```diff
@@ -29,6 +29,133 @@ resource "kubernetes_namespace" "this" {
...
+        port = "metrics"
+        protocol = "TCP"
+      }
+    }
+  }
+}
+
+resource "kubernetes_network_policy" "allow_ingress_traffic" {
+  metadata {
+    name      = "allow-ingress-traffic"
```
simongottschlag, line 117: Are we sure network policies are applied to LB services? Is it the same in all clouds or different?

phillebaba: Network policies have no effect on services, only on pods. What we are doing is allowing traffic originating from IPs outside the cluster, to reach the nginx pods.

modules/kubernetes/ingress-nginx/main.tf
phillebaba: Yes we should add some flag to flip the two CIDRs with a flag if it is a private ingress controller.

modules/kubernetes/ingress-healthz/main.tf
phillebaba: Not sure if this is something currently used by kube-dns so I need to look further into it.

simongottschlag: @phillebaba it should be used by all modern DNS servers, or else larger lookups (local or external) wouldn't work. I've witnessed it myself not working as expected with some versions of alpine.

If you want to check the code, kube-dns is using skydns: https://github.com/skynetservices/skydns/blob/master/server/server.go

modules/kubernetes/ingress-nginx/main.tf
phillebaba: We had to remove this as internal traffic hitting the ingress controller shows up as coming from an internal IP.

modules/kubernetes/ingress-healthz/main.tf
```diff
@@ -31,6 +31,88 @@ resource "kubernetes_namespace" "this" {
...
+        "app.kubernetes.io/instance" = "ingress-healthz"
+        "app.kubernetes.io/name" = "nginx"
+      }
+    }
+
+    policy_types = ["Ingress"]
+
+    ingress {
+      ports {
+        port = "8080"
```
NissesSenap, line 95: Instead of using port number it's not worth to use name?
I played around with the editor.cilium.io and apparently it dosen't support strings but in reality it's is.

```shell
 k explain networkpolicy.spec.ingress.ports
   port	<string>
     The port on the given protocol. This can either be a numerical or named
     port on a pod. If this field is not provided, this matches all port names
     and numbers.
```

NissesSenap:
You might already know this but apparently there are two ways AKS talks to Azure.
One way is `tunnelfront`and the other one is called `aks-link` I don't think we need to take this in to consideration since the network policies is already in place for them but good to think about.

PR #345, opened 2021-07-12
Add support for AWS in Xenit proxy

phillebaba:


simongottschlag:
@phillebaba there seems to be a lock-file that has been commited

NissesSenap (approved):
LGTM

PR #346, opened 2021-07-12
Enable prometheus in eks-core

NissesSenap:


phillebaba (dismissed):

NissesSenap:
I can't decide if I like the solution of adding metrics server in the prometheus-extras helm chart.
In some ways it's a very natural location. But we will be editing kube-system from a helm chart that primarily focuses on prometheus even if the pod/servicemonitors is outside of the namespace. 

Since this is closely related to monitoring I see this as the most natural way forward. What do you think @simongottschlag @phillebaba ?

phillebaba:
I will need to dig deeper into the differences between AKS and EKS, but I agree with you that adding things to kube-system is not an optimal solution.

NissesSenap:
You can read more here: https://aws.amazon.com/premiumsupport/knowledge-center/eks-metrics-server/

simongottschlag:
@NissesSenap I don't understand why we need to add it to `kube-system`. Is there some kind of requirement in regards to this?

simongottschlag:
@NissesSenap if that's the case, I think the most logical step would be to create a new module for it that installs it in `kube-system`.

NissesSenap:
There is also a helm chart available https://github.com/helm/charts/tree/master/stable/metrics-server isn't super updated but that doesn't mean there is anything wrong with it.
With my last commit I got it to work as well.
So now I can perform `kubectl top node` :) 

I will start to create a new k8s module after lunch.

phillebaba (approved):

simongottschlag:
@NissesSenap nice! Bitnami seems to be more up to date.

PR #349, opened 2021-07-15
aws vpc peering config

NissesSenap:


modules/aws/core/vpc.tf
```diff
@@ -166,29 +198,73 @@ resource "aws_route_table_association" "private" {
 # VPC Peering
 resource "aws_vpc_peering_connection" "this" {
```
NissesSenap, line 199: Instead of this should I call it requester?

modules/aws/core/variables.tf
```diff
@@ -31,22 +38,30 @@ variable "vpc_config" {
   })
 }
 
-variable "vpc_peering_enabled" {
-  description = "If true vpc peering will be configured"
-  type        = bool
-  default     = true
+variable "vpc_peering_config_requester" {
+  description = "VPC Peering configuration"
```
phillebaba, line 42: Update the description.

modules/aws/core/variables.tf
```diff
@@ -31,22 +38,30 @@ variable "vpc_config" {
...
+    peer_owner_id          = string
+    peer_vpc_id            = string
+    destination_cidr_block = string
+  }))
+  default = []
 }
 
-variable "vpc_peering_config" {
+variable "vpc_peering_config_accepter" {
   description = "VPC Peering configuration"
```
phillebaba, line 53: Ditto

modules/aws/core/variables.tf
```diff
@@ -31,22 +38,30 @@ variable "vpc_config" {
...
-    destination_cidr_block = ""
-    peer_owner_id          = ""
-    peer_vpc_id            = ""
-  }
+  }))
+  default = []
+}
+
+
+variable "internet_gateway_enabled" {
```
phillebaba, line 63: Do we need this if we can just check the count of the public subnets?

phillebaba (approved):

phillebaba (approved):
LGTM

PR #355, opened 2021-07-20
Add fluxcd EKS core

NissesSenap:
Creating a bad PR where I also do a change to AZAD by adding a annotation to it.
I also bump the AZAD helm version


simongottschlag (approved):
LGTM, but would prefer the azad-kube-proxy changes in separate PRs

NissesSenap:
Me too but to lazy right now.

PR #357, opened 2021-07-23
Enable to create a private ingress class

NissesSenap:
This way customers that don't want to make endpoints public on the internet can do so.


NissesSenap:
Currently hitting allot of  hostNetwork issues.
It's not a problem to do what I do with a single instance running in the cluster but if i run a private and a public he pods don't want to start due to already used ports.

Don't think it's to many people trying to run hostNetwork at the same time on multiple clusters ^^
I GCP they have to open extra ports to be able forward data to the new internal ports. That might be true with this config as well, we will have to try.

But before that I hav to find the perfect helm chart config where I manage to overwrite all current ports in a good way.

If you have any other idea please share.

NissesSenap:
A number of tries later I have gotten the nginx controller to start when running hostPort.
Now I will try to verify that it actually works as well anod not just starts.

NissesSenap:
Have verified that it works, I'm able to have two ingress classes and I have tried using two different `classNames` in the ingress yaml and the controllers picks it up without a problem.

Worth noticing is that the current solution create 2 ingressClass, for example with the name:

```shell
k get ingressclasses          
NAME            CONTROLLER             PARAMETERS   AGE
nginx           k8s.io/ingress-nginx   <none>       76m
nginx-private   k8s.io/ingress-nginx   <none>       76m
```

I make it possible to set a default ingressClass on the private one but not on the public, since I don't think it's a good idea.

If we want we could make it possible to set name_override on the **public** ingressclass as well but I would have to write rather ugly if statments in values.yaml.
How important do you think it is @phillebaba ?

Currently we don't support adding more then two ingresses using the nginx-ingress controller. If you have some other controller in your cluster it shouldn't be a problem.

NissesSenap:
I have verified on azure with a single setup and this PR won't create any changes to it just as intended.
As long as we don't think that it's important in a dual mode that the ingressclass should be called
```
nginx-public
nginx-private
```
and it's okay with the current solution of

```
nginx
nginx-private
``` 

Then I think we should be okay to merge this.

modules/kubernetes/ingress-nginx/main.tf
```diff
@@ -22,33 +22,100 @@ terraform {
...
+    default_certificate = {
+      enabled         = var.default_certificate.enabled
+      dns_zone        = var.default_certificate.dns_zone
+      namespaced_name = "ingress-nginx/ingress-nginx"
+    }
+    extra_config          = var.extra_config
+    extra_headers         = var.extra_headers
+    linkerd_enabled       = var.linkerd_enabled
+    multiple_ingress      = false
+    default_ingress_class = false
```
phillebaba, line 58: This should probably be set to true even if it the class is nginx.

simongottschlag: removed `multiple_ingress` variable

PR #360, opened 2021-07-28
EKS use private nodes together with bastion for breakglass solution

NissesSenap:
Send in a list of CIDR that is avaliable to talk to AWS EKS endpoint


NissesSenap:
Will move over to using private only.
Will keep on working in this PR.

simongottschlag:
Closing in favor of #365

PR #361, opened 2021-07-29
Add calico-system to ignore list

NissesSenap:
During a upgrade to AKS 1.20 calcio-system starts to getting used also in AKS.


NissesSenap:
Ignore the downgrade of the OPA helm chart, it's just for my test. I will update this later

dachrillz (approved):

PR #363, opened 2021-07-30
Add linkerd initContainer

NissesSenap:
This to solve issues with linkerd not starting on a node where the CNI is fully online.
See: https://github.com/linkerd/linkerd2/issues/2219

### DIFF
--- a/modules/kubernetes/linkerd/main.tf
+++ b/modules/kubernetes/linkerd/main.tf
@@ -192,12 +192,14 @@ resource "kubernetes_secret" "webhook_issuer_tls" {
 }
 
 # Install linkerd-cni helm chart
+# Tmp use edge version until 2.11.0 is released where we will get the helm chart feature we need.
+#tf-latest-version:ignore
 resource "helm_release" "linkerd_cni" {
-  repository = "https://helm.linkerd.io/stable"
+  repository = "https://helm.linkerd.io/edge"
   chart      = "linkerd2-cni"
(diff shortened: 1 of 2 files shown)


NissesSenap:
So this won't create any difference due to the PR with this functionality isn't part of this helm chart for some reason.

I posted a question about this in the linkerd slack

```
This PR was merged the 15th of april and the
stable 2.10.2 release came out the 2021-05-18 according to https://helm.linkerd.io/stable/index.yaml
But still this PR isn't in the helm chart.
Would it be possible the generate a new helm release version?
``` 

And of course by pure luck when I started linkerd against this commit it worked...
But from reading the other issues on the linkerd page It seems to me like everything was already done on the node level and that was why it worked.

Let's see if we can get a new helm chart out, until then there is no point in merging this.

simongottschlag (approved):
LGTM 😊👍🥇

NissesSenap:
In 2.10.2 the initContainer feature that we needed wasn't released.
To solve this we have temporarily started to use the linkerd-edge release.

We won't follow along with the releases so that is why we use the `#tf-latest-version:ignore` tag.

Tried to use the new helm chart but with the old container image but in the next release linkerd will be able to run rootless so the helm chart is using a number of temporary folders that 2.10.2 is built to handle so I had to use the latest container image as well.

In our monitoring I haven't seen any issues so far so we decided to merge this PR.

PR #365, opened 2021-08-05
Enable CIDR definition in EKS clusters

NissesSenap:
Enable private endpoints on the EKS cluster,
this way the cluster can communicate inside the VPC when communicating internally.


NissesSenap:
This is much easier then #360 it allows us to define a limited CIDR scope and enable local endpoints but don't mix in any bastions or similar solutions. This is how we have it in AKS today.

One of them will only be merged.

simongottschlag (approved):
LGTM

phillebaba (approved):

PR #371, opened 2021-08-17
Change linkerd proxy default log format

phillebaba:
This is a small change which sets the linkerd proxy to by default output
logs as json instead of plain text. The reasoning is that log collectors
such as datadog prefer structured log data so the majority of users
prefer this behavior instead of plain text. It is still possible to
change the format to plain text with an annotation on the pod.

### DIFF
--- a/modules/kubernetes/linkerd/templates/values.yaml.tpl
+++ b/modules/kubernetes/linkerd/templates/values.yaml.tpl
@@ -40,6 +40,8 @@ enablePodAntiAffinity: true
 
 # proxy configuration
 proxy:
+  # A better default for log collectors that require strucutured data
+  logFormat: json
   resources:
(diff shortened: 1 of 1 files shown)


simongottschlag (approved):
LGTM

modules/kubernetes/linkerd/templates/values.yaml.tpl
```diff
@@ -40,6 +40,8 @@ enablePodAntiAffinity: true
 
 # proxy configuration
 proxy:
+  # A better default for log collectors that require strucutured data
```
anderscarling, line 43: Supernitpicky detail: `structured` is misspelled 🙂

simongottschlag (approved):

PR #373, opened 2021-08-18
Fix validation

simongottschlag:
`make validate` doesn't error if directory doesn't exist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,23 +68,22 @@ tfsec:
 validate:
 	set -e
 
-	tf_validate () {(set -e
-		cd $$1
-		rm -f .terraform.lock.hcl
-		terraform init 1>/dev/null
-		set +e
-		OUTPUT=$$(terraform validate 2>&1)
-		EXIT_CODE=$$?
+	tf_validate () {
+		rm -f $$1/.terraform.lock.hcl
+		terraform -chdir=$$1 init 1>/dev/null 2>&1
+		TEMP_FILE=$$(mktemp)
+		
 		set -e
-		if [[ $$EXIT_CODE -ne 0 ]]; then
-			echo terraform-validate: $$1 failed
-			echo "$$OUTPUT"
-			exit $$EXIT_CODE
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -68,7 +68,7 @@ tfsec:
 validate:
 	set -e
 
-	tf_validate () {
+	tf_validate () {(set -e
```
phillebaba, line 71: NIT: Could you have this on its own line.

phillebaba (dismissed):

Makefile
```diff
@@ -68,7 +68,7 @@ tfsec:
...
 		cd $$1
```
bittrance, line 72: `cd` mutates a global variable and should be avoided as much as possible.

Makefile
```diff
@@ -84,7 +84,7 @@ validate:
 
 		echo terraform-validate: $$1 succeeded
 		exit $$EXIT_CODE
-	}
+	)}
```
bittrance, line 87: Do you really need a subshell? Functions will return the exit code from the last line and a script will terminate if a function returns non-zero if `set -e`. Generally speaking, bash script should always start with `set -eu` (and probably `set -o pipefail`) and should not have to turn any of them off.

validation/kubernetes/cluster-autoscaler/main.tf
```diff
@@ -0,0 +1,21 @@
...
+
+module "cluster_autoscaler" {
+  source = "../../../modules/kubernetes/cluster-autoscaler"
+
+  providers = {
+    kubernetes = kubernetes
+    helm       = helm
+  }
+
+  cluster_name   = "foo"
```
bittrance, line 15: Nitpick: I suspect the output will be more readable if this would be `test-cluster` or `some-cluster` or `ze-cluster` or something.

bittrance (commented):
I think the `tf_validate` function could be reformulated to do what you want and be both shorter and more bashy. I will be happy to do some pair programming to find out what the best style would be.

bittrance (approved):

PR #380, opened 2021-08-25
aws: change from elb to nlb

simongottschlag:
This PR changes the default behaviour for ingress-nginx in EKS to use NLB instead of ELB. This to lower cost and have a behaviour more like Azure Load Balancer and how it is possible to forward the client ip (real ip) without proxy protocol.

This also fixes the if statements in the template values to remove unneccessary amount of blank lines.


simongottschlag:
@phillebaba ready for review, now working and nginx is seeing the real ip.

PR #383, opened 2021-08-27
Add AWS style guide

phillebaba:
This change aims to document an AWS specific style guide and implement it in all AWS resources.

### DIFF
--- a/modules/aws/README.md
+++ b/modules/aws/README.md
@@ -14,8 +14,8 @@ This directory contains all the AWS Terraform modules.
 This section covers AWS specific standards that all modules shoudl try to follow to decrease the risk of annoying problems in the future, and to guide
 new contributors in how resources should be named. Any changes to this guide should be done together with updating the Terraform.
 
-
 ### Names and tags
+
 A quirk of AWS is that there are some resources that are global and other that are regional. Ontop of that there are resources that get automatically
 generated IDs and others that use the given name as a unique ID. That name may either have to be unique within the account, within that AWS region or
 across all AWS regions.
@@ -23,14 +23,14 @@ across all AWS regions.
 Global resources with unique names:
 * IAM Roles
 * IAM Policy
+* S3 Bucket
 
 Regional resources with unique names:
 * EKS Clusters
-* S3 Bucket
 
 Any resource that has a generated ID should have tags to distinguish the resources between each other. The following tags should be added to all
-resources of this kind. If multiple instances of the same resource is expected to be deployed, such as EKS, the name of the resource should include a
-unique suffix to distinguish the resources.
+resources of this kind. Try to make sure that the values are unique enough that there is no overlap.
+
 | Tag | Required | Description |
 | --- | --- | --- |
 | Name | Yes | Name of the resource, try to convey the purpose or use. |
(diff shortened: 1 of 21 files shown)


modules/aws/core/route53.tf
```diff
@@ -7,8 +7,10 @@ resource "aws_route53_zone" "this" {
...
-  tags = {
-    Name        = var.dns_zone
-    Environment = var.environment
-  }
+  tags = merge(
+    local.global_tags,
+    {
+      Name        = var.dns_zone
+    },
+  )
```
simongottschlag, line 15: Any point in using default tags? https://registry.terraform.io/providers/hashicorp/aws/latest/docs#default_tags

modules/aws/README.md
```diff
@@ -8,3 +8,51 @@ This directory contains all the AWS Terraform modules.
...
+* S3 Bucket
+
+Any resource that has a generated ID should have tags to distinguish the resources between each other. The following tags should be added to all
+resources of this kind. If multiple instances of the same resource is expected to be deployed, such as EKS, the name of the resource should include a
+unique suffix to distinguish the resources.
+| Tag | Required | Description |
+| --- | --- | --- |
+| Name | Yes | Name of the resource, try to convey the purpose or use. |
+| Environment | Yes | Environment the resource is deployed in. |
+| Module | Yes | The name of the Terraform module which has created the resource. |
```
bittrance, line 38: Is this not rather the module that defined the resources? It may well because of configuration in a downstream module.

bittrance (dismissed):
I think this is reasonable. Minor comments below.

Spelling:
- ontop -> on top
- shoudl -> should

modules/aws/core/route53.tf
phillebaba: Default tags only work in the provider so I would rather not.

modules/aws/README.md
```diff
@@ -8,3 +8,51 @@ This directory contains all the AWS Terraform modules.
...
+generated IDs and others that use the given name as a unique ID. That name may either have to be unique within the account, within that AWS region or
+across all AWS regions.
+
+Global resources with unique names:
+* IAM Roles
+* IAM Policy
+
+Regional resources with unique names:
+* EKS Clusters
+* S3 Bucket
```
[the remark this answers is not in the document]
phillebaba: Yes you are right, fixed this.

modules/aws/README.md
```diff
@@ -8,3 +8,51 @@ This directory contains all the AWS Terraform modules.
...
+## Style Guide
+
+This section covers AWS specific standards that all modules shoudl try to follow to decrease the risk of annoying problems in the future, and to guide
+new contributors in how resources should be named. Any changes to this guide should be done together with updating the Terraform.
+
+
+### Names and tags
+A quirk of AWS is that there are some resources that are global and other that are regional. Ontop of that there are resources that get automatically
+generated IDs and others that use the given name as a unique ID. That name may either have to be unique within the account, within that AWS region or
+across all AWS regions.
```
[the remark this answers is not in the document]
phillebaba: I removed one of them to align with the S3 unique name change.

modules/aws/README.md
```diff
@@ -8,3 +8,51 @@ This directory contains all the AWS Terraform modules.
...
+Any resource that has a generated ID should have tags to distinguish the resources between each other. The following tags should be added to all
+resources of this kind. If multiple instances of the same resource is expected to be deployed, such as EKS, the name of the resource should include a
+unique suffix to distinguish the resources.
+| Tag | Required | Description |
+| --- | --- | --- |
+| Name | Yes | Name of the resource, try to convey the purpose or use. |
+| Environment | Yes | Environment the resource is deployed in. |
+| Module | Yes | The name of the Terraform module which has created the resource. |
+| Cluster | No | If the resources is specific to an EKS cluster the cluster name should be entered here. |
+| Application | No | Name of the application the resource may be tightly linked to. |
```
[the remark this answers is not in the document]
phillebaba: These tags are for us in our own accounts. We will never (hopefully) deploy these modules in shared accounts, so conflict with other users should not happen.

modules/aws/README.md
phillebaba: We should not have nested modules except in very specific circumstances. The only nested module we have today is IRSA which is according to me is useful. I think it is easiest to have the first module name in the tag.

NissesSenap (approved):
I think it looks good, just one more make docs and we should be set.
Great job!

NissesSenap (approved):
LGTM

PR #385, opened 2021-08-30
Add datadog_enabled to ingress-nginx

NissesSenap:
Thanks to this datadog will be able to get metrics from nginx

### DIFF
--- a/modules/kubernetes/ingress-nginx/templates/values.yaml.tpl
+++ b/modules/kubernetes/ingress-nginx/templates/values.yaml.tpl
@@ -50,17 +50,19 @@ controller:
     default-ssl-certificate: "${default_certificate.namespaced_name}"
   %{~ endif ~}
 
+  %{~ if datadog_enabled || linkerd_enabled ~}
   podAnnotations:
-  %{~ if datadog_enabled ~}
+    %{~ if datadog_enabled ~}
     ad.datadoghq.com/controller.check_names: '["nginx_ingress_controller"]'
     ad.datadoghq.com/controller.init_configs: '[{}]'
     ad.datadoghq.com/controller.instances: '[{"prometheus_url": "http://%%host%%:10254/metrics"}]'
-  %{~ endif ~}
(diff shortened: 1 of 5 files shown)


NissesSenap:
With the following annotation to the nginx pods:

```.yaml
  annotations:
    ad.datadoghq.com/controller.check_names: '["nginx_ingress_controller"]'
    ad.datadoghq.com/controller.init_configs: '[{}]'
    ad.datadoghq.com/controller.instances: '[{"prometheus_url": "http://%%host%%:10254/metrics"}]'
```

and I can see the data in datadog

image

I need to verify what happens if `podAnnotations` is left empty without datadog nor linkerd is used...

I also think that it might be interesting to get nginx data to datadog, right now we only send ingress-controller data.
For more documentation see: https://docs.datadoghq.com/integrations/nginx_ingress_controller/?tab=containerized

I will continue on this tomorrow.

modules/kubernetes/ingress-nginx/templates/values.yaml.tpl
```diff
@@ -50,8 +50,13 @@ controller:
     default-ssl-certificate: "${default_certificate.namespaced_name}"
   %{~ endif ~}
 
-  %{~ if linkerd_enabled ~}
```
simongottschlag, line 53: I would prefer it if we keep this if statement and adding `|| datadog_enabled`

simongottschlag (changes requested):

NissesSenap:
After a discussion with @simongottschlag we think that the data that we get from nginx-ingress-controller /metrics page should be enough to get the data that we need. Sadly the datadog docs isn't up to data so it sent me on a longer configuration trip then I have hoped for.
Any way please review again and hopefully this should be enough.

simongottschlag (approved):
LGTM 😊👍🥇

NissesSenap:
@simongottschlag can you take a second look? 
I missed the azure addition in my last PR.

simongottschlag (approved):

PR #388, opened 2021-08-31
Enable clusterChecksEnabled in datadog

NissesSenap:
* Enable metrics to be gatherd about the pods, like number of replicas.
* Set replicas: 2 in the clusterAgent deployment

### DIFF
--- a/modules/kubernetes/datadog/charts/datadog-extras/templates/datadogagent.yaml
+++ b/modules/kubernetes/datadog/charts/datadog-extras/templates/datadogagent.yaml
@@ -48,6 +48,9 @@ spec:
     priorityClassName: platform-low
     image:
       name: "gcr.io/datadoghq/cluster-agent:latest"
+    config:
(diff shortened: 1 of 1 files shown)


simongottschlag (approved):
LGTM 🙌

NissesSenap:
Getting errors like:
```shell
2021-08-31 14:14:01 UTC | CLUSTER | WARN | (pkg/clusteragent/clusterchecks/dispatcher_main.go:136 in add) | No available node to dispatch kubernetes_state_core:e7dac77010842ea4 on, will retry later
2021-08-31 14:14:04 UTC | CLUSTER | INFO | (pkg/api/healthprobe/healthprobe.go:73 in healthHandler) | Healthcheck failed on: [clusterchecks-dispatch]
```

Feels like I'm missing some agent config to get this to work...

NissesSenap:
I don't think this is needed, it seems like we are getting all the data we should now with the updated operator. I will close this.

PR #397, opened 2021-09-13
Update specific providers

phillebaba:
This PR updates specific providers until we can ignore those which cant be updated.
* hashicorp/aws
* hashicorp/azurerm
* hashicorp/helm
* integrations/github
* fluxcd/flux


NissesSenap (approved):
LGTM

PR #404, opened 2021-09-21
Disable OPA in prometheus ns in EKS

NissesSenap:
For some reason the mutation don't seem to trigger for the statefulset
created by the prometheus operator.


bittrance:
With line breaks and sorted alphabetically?

dachrillz (approved):

Issue #378, opened 2021-08-23
falco not working with eks


simongottschlag:
When using falco with eks, the following is logged:
```
* Setting up /usr/src links from host
* Running falco-driver-loader for: falco version=0.29.1, driver version=17f5df52a7d9ed6bb12d3b1768460def8439936d
* Running falco-driver-loader with: driver=bpf, compile=yes, download=yes
* Mounting debugfs
* Trying to download a prebuilt eBPF probe from https://download.falco.org/driver/17f5df52a7d9ed6bb12d3b1768460def8439936d/falco_amazonlinux2_5.4.117-58.216.amzn2.x86_64_1.o
curl: (22) The requested URL returned error: 404 
Unable to find a prebuilt falco eBPF probe
* Trying to compile the eBPF probe (falco_amazonlinux2_5.4.117-58.216.amzn2.x86_64_1.o)
make[1]: *** /lib/modules/5.4.117-58.216.amzn2.x86_64/build: No such file or directory.  Stop.
make: *** [Makefile:18: all] Error 2
mv: cannot stat '/usr/src/falco-17f5df52a7d9ed6bb12d3b1768460def8439936d/bpf/probe.o': No such file or directory
Unable to load the falco eBPF probe
Mon Aug 23 16:12:07 2021: Falco version 0.29.1 (driver version 17f5df52a7d9ed6bb12d3b1768460def8439936d)
Mon Aug 23 16:12:07 2021: Falco initialized with configuration file /etc/falco/falco.yaml
Mon Aug 23 16:12:07 2021: Loading rules from file /etc/falco/falco_rules.yaml:
Mon Aug 23 16:12:08 2021: Loading rules from file /etc/falco/falco_rules.local.yaml:
Mon Aug 23 16:12:08 2021: Loading rules from file /etc/falco/rules.d/rules_user_known_k8s_api_callers.yaml:
Mon Aug 23 16:12:08 2021: Loading rules from file /etc/falco/rules.d/rules_user_known_k8s_client_container.yaml:
Mon Aug 23 16:12:08 2021: Loading rules from file /etc/falco/rules.d/rules_user_known_write_below_etc_activities.yaml:
Mon Aug 23 16:12:09 2021: Loading rules from file /etc/falco/rules.d/rules_user_sensitive_mount_containers.yaml:
Mon Aug 23 16:12:09 2021: Unable to load the driver.
[...]

PR #405, opened 2021-09-21
Disable eBPF on AWS

NissesSenap:
Solves #378, sadly amazon linux kernel don't seem to be new enough to use eBPF but the pre-built kernel should work just fine.


NissesSenap:
The container starts if docker is disabled but it dosen't give me all the data that I want. It's unable to pinpoint the container image that is used to perform the call and thus any of our ignore image rules get's ignored.

simongottschlag (approved):
LGTM!

PR #408, opened 2021-09-23
Add aquasec starboard to xkf

NissesSenap:
Install a seperate trivy instance on the cluster to store the
vulnerability database.

### DIFF
--- a/validation/kubernetes/starboard/main.tf
+++ b/validation/kubernetes/starboard/main.tf
@@ -1,0 +1,16 @@
+terraform {}
+
+provider "kubernetes" {}
+
+provider "helm" {}
+
+module "falco" {
+  source = "../../../modules/kubernetes/starboard"
+
+  providers = {
+    kubernetes = kubernetes
+    helm       = helm
+  }
+
+  cloud_provider = "bar"
+}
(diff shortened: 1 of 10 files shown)


NissesSenap:
Starboard get's ignored in OPA due to CIS benchmark getting run on each node.

NissesSenap:
How do we want to handle EKS?
https://aquasecurity.github.io/starboard/v0.12.0/integrations/managed-registries/
We need to create a IAM rule to be able to pull images stored in ECR, this isn't a problem in azure.

Should we do this in the kubernetes module?
Or should the users provide a IAM rule to use in the SA that is needed?

What do you think @bittrance @phillebaba @simongottschlag

bittrance:
Since the ECR repo is likely to be hosted in another AWS account, I think it will be tricky to create the IAM you need. If there is no way to scan images on ECR, I think that will have to be a separate PR. If it is easy, you can always have a variable to pass in the role/policy ARN to attach to a created ServiceAccount, assuming there is support for assuming cross-account roles.

validation/kubernetes/starboard/main.tf
```diff
@@ -0,0 +1,16 @@
+terraform {}
+
+provider "kubernetes" {}
+
+provider "helm" {}
+
+module "falco" {
```
simongottschlag, line 7: Wrong name

phillebaba (approved):

PR #409, opened 2021-09-29
Automatically close stale issues and PRs

### LINKED ISSUE #119: GH Action to close stale issues/PRs
https://github.com/marketplace/actions/close-stale-issues

bittrance:
GitHub action to mark issues and PRs as stale after 90 days on inactivity unless they are labelled `accepted`.

Fixes #119.

### DIFF
--- a/.github/workflows/stale.yaml
+++ b/.github/workflows/stale.yaml
@@ -7,7 +7,7 @@ name: Mark stale issues and pull requests
 
 on:
   schedule:
-  - cron: '0 4 * * *'
+    - cron: '0 4 * * *'
 
 jobs:
   stale:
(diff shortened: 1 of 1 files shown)


simongottschlag (dismissed):
LGTM 😊👍

.github/workflows/stale.yaml
```diff
@@ -0,0 +1,36 @@
...
+        stale-pr-message: >
+          This PR hasn't been updated for a while, marking as stale.
+        stale-issue-label: stale
+        stale-pr-label: stale
+        days-before-stale: 90
+        days-before-close: 7
+        exempt-issue-labels: accepted
+        exempt-pr-labels: accepted
+        remove-issue-stale-when-updated: true
+        remove-pr-stale-when-updated: true
```
NissesSenap, line 36: Please add enter on line 36. Git diff is not great a handling files without line breaks at the end.

.github/workflows/stale.yaml
```diff
@@ -0,0 +1,36 @@
+# This workflow warns and then closes issues and PRs that have had no 
+# activity for a specified amount of time.
+#
+# You can adjust the behavior by modifying this file.
+# For more information, see: https://github.com/actions/stale
+name: Mark stale issues and pull requests
+
+on:
+  schedule:
+  - cron: '0 4 * * *'
```
NissesSenap, line 10: I think we can increase how often this job is run. It would make me kind of sad updating a issue/pr and for the entire day it keeps saying stale.

.github/workflows/stale.yaml
```diff
@@ -0,0 +1,36 @@
...
+jobs:
+  stale:
+
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+
+    steps:
+    - uses: actions/stale@v3
```
NissesSenap, line 21: Update to `@v4`

NissesSenap (changes requested):
Please have a look at comments.
I also think we should have a internal discussion on how to manage our issues before merging this. I think most of our issues shouldn't be deleted after 90 days since I still think they are relevant even if we haven't prioritized them and I don't want to become a bot fighting the bot writing comments in a bunch of issues so I would use the label accepted on most issues that we have today.

.github/workflows/stale.yaml
bittrance: That is true, but isn't that best addressed by also running the workflow when there are comments, commits, et c?

.github/workflows/stale.yaml
bittrance: Fixed.

.github/workflows/stale.yaml
bittrance: Fixed.

bittrance:
> I also think we should have a internal discussion on how to manage our issues before merging this. 

I assume you mean that a PR discussion is insufficient. What sort of forum would you prefer?

> I think most of our issues shouldn't be deleted after 90 days since I still think they are relevant ...

The PRs and issues are not deleted tho? They are simply closed. Given that you should always search existing issues before creating a new one, I'm not sure I see how closing them is a big thing?

.github/workflows/stale.yaml
NissesSenap: Yeah we can do that as well, but if we are using that I guess we can disable the cronjob completely. If we trigger the job on any comment it will check all the issue not just the issue that we commented on.

PR #411, opened 2021-10-04
Bump azad-kube-proxy to enable separate port for metrics and health checks

simongottschlag:
This PR bumps the version of azad-kube-proxy from `v0.0.24` to `v0.0.26` where the metrics and health checks are now running on a separate port (https://github.com/XenitAB/azad-kube-proxy/pull/298).

It also removes a duplicate environment variable of PORT which is already set by `application.port`.


NissesSenap (approved):
LGTM

PR #413, opened 2021-10-07
Add aws flow log for easy deploy

nissessenap:
A easy way to setup flow logs to debug VPC traffic.
Disabled by default


bittrance:
This module looks like a good thing for debugging network, but Cloudwatch Logs is rather expensive, so at scale you prolly want to send output to S3 (and you prolly want Athena to analyze it too). I like that this module is narrow: perhaps it should be called "debug-vpc-networking" or "flow-logs-to-cloudwatch" or something?

Issue #422, opened 2021-10-20
Create CI that checks if CHANGELOG.md have been updated in a PR


nissessenap:
Overall we should always update CHANGELOG.md in our PR:s when creating a PR for this repo.

It should trigger on each PR
From a simple bash command we should be able to verify if the file have changed

```shell
git diff --exit-code CHANGELOG.md
```

But there will be use cases where we don't want to be forced to update CHNAGELOG.md, to solve this my suggestion is to use a label on the PR. If the PR is set it should automatically ignore the PR, ether don't run the PR check at all or the script should have some logick that checks the Github API if the label is set for a specific PR. This will of course require a github secret.

There is probably a simple github action available that will download the GH cli for you https://github.com/cli/cli where you can check for labels in PR:s.

PR #424, opened 2021-10-20
CI check if CHANGELOG have been updated

nissessenap:
Relates to #422, if you don't want to run this check you can add the label
ignore-changelog and it will be okay to not provide a changelog update in your PR.


NissesSenap:
Before changing the CHANGELOG.md i will now get:

image

PR #433, opened 2021-10-26
Add CRD installation to modules

phillebaba:


phillebaba:
@bittrance we will have to pause this PR for a while as I found a bug in the Helm Terraform Provider. I have submitted PR hashicorp/terraform-provider-helm/pull/795 that fixes these issues. We will have to wait for it to be merged and released before progressing with this PR, as it would cause more harm than good before that.

NissesSenap:
I guess we can close this PR and instead look at: https://github.com/XenitAB/terraform-modules/pull/664

PR #434, opened 2021-10-28
Add config for AKS automatic_channel_upgrade & upgrade_settings

CalleB3:
Adding configuration for AKS 
automatic_channel_upgrade & upgrade_settings configuration in terraform-modules.
Implementing to simplify and automate the updating / upgrading process for K8s clusters.


NissesSenap:
@simongottschlag @phillebaba do you have any thoughts on this one?
I think this PR should do the trick.

CalleB3:
To be backwards compatible and able to run without "automatic_channel_upgrade" set variable to "null".
 "aks_upgrade_channel = null"

Here is the different channels you can use.
https://docs.microsoft.com/en-us/azure/aks/upgrade-cluster#set-auto-upgrade-channel-preview
Could recommend using "node-image" to patch the nodes automatically.

NissesSenap (approved):

ahockersten (approved):
LGTM

phillebaba:
The max surge setting is a good option, as right now it will do a single node at a time. 

I am not too convinced about the automatic upgrade channel. My personal opinion is that we should not enable it. But if we are going to enable it, then it should not be a variable but a hard coded setting that we use everywhere. I do not want the possibility of different behaviors in different clusters.

NissesSenap:
Me and @phillebaba discussed a bit and setting "node-image" as the default value could be an option.

@CalleB3 is there any information from azure how these node-images gets updated on the nodes?
Preferably we would want to make sure that all the clusters don't start to changing nodes at the same times.
Even better would be if we could get them to split it with a certain delay in prod after all of dev is done, like a few days.

I don't think these options exist now but do they have some docs that might explain how "node-image" works?

CalleB3:
> Me and @phillebaba discussed a bit and setting "node-image" as the default value could be an option.
> 
> @CalleB3 is there any information from azure how these node-images gets updated on the nodes? Preferably we would want to make sure that all the clusters don't start to changing nodes at the same times. Even better would be if we could get them to split it with a certain delay in prod after all of dev is done, like a few days.
> 
> I don't think these options exist now but do they have some docs that might explain how "node-image" works?

I found this:
https://docs.microsoft.com/en-us/azure/aks/planned-maintenance
&
https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/maintenance_configuration
Where you can use automatic upgrade and only allow them to perform during set maintenance windows.
Not entirely sure on how the config would look but should be considerable.

NissesSenap:
Found https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster#maintenance_window I think this sounds like a good option. I assume that this would apply to all the nodes in the cluster but I don't know. We would have to try this out.
@phillebaba do you think this could be a possible way forward?

phillebaba:
I still dont really agree with this. I would not be comfortable doing this in the middle of the night automatically right now. Then again if everyone else wants to I am not going to block it.

NissesSenap:
I see your point, this is a feature that we can add in the future.
For now we close it. Thanks for your work on this @CalleB3

PR #437, opened 2021-11-03
Add podmonitor for secrets-store-csi-driver

### LINKED ISSUE #356: Add podmonitor for secrets-store-csi-driver
Today we monitor csi-secrets-store-provider-azure but not the csi DS it self.
When: https://github.com/kubernetes-sigs/secrets-store-csi-driver/pull/614 get's merged we should be able to do so as well.

Update: https://github.com/XenitAB/terraform-modules/blob/main/modules/kubernetes/prometheus/charts/prometheus-extras/templates/monitor.yaml

CalleB3:
Closes #356

Enables podmonitor for Prometheus to monitor "secrets-store-csi-driver" for both Azure and AWS.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## Unreleased
 
+
+### Added
+
+- [#437](https://github.com/XenitAB/terraform-modules/pull/437) Add podmonitor for secrets-store-csi-driver
+
 ## 2021.11.6
 
 ### Added
--- a/modules/kubernetes/prometheus/charts/prometheus-extras/templates/monitor.yaml
+++ b/modules/kubernetes/prometheus/charts/prometheus-extras/templates/monitor.yaml
@@ -312,3 +312,37 @@ spec:
     - path: /metrics
       port: metrics
 {{- end }}
+{{- if .Values.enabledMonitors.csiSecretsStorProviderAzure }}
+apiVersion: monitoring.coreos.com/v1
+kind: PodMonitor
+metadata:
+  name: csi-secrets-store-driver-azure
+  namespace: csi-secrets-store-provider-azure
+  labels:
+    xkf.xenit.io/monitoring: platform
+spec:
+  selector:
+    matchLabels:
+      app.kubernetes.io/instance: csi-secrets-store-provider-azure
+      app.kubernetes.io/name: csi-secrets-store-provider-azure
+  podMetricsEndpoints:
+    - path: /metrics
+      port: metrics
+{{- end }}
+{{- if .Values.enabledMonitors.csiSecretsStorProviderAws }}
+apiVersion: monitoring.coreos.com/v1
+kind: PodMonitor
+metadata:
+  name: csi-secrets-store-driver-aws
+  namespace: csi-secrets-store-provider-aws
+  labels:
+    xkf.xenit.io/monitoring: platform
+spec:
+  selector:
+    matchLabels:
+      app.kubernetes.io/instance: csi-secrets-store-provider-aws
+      app.kubernetes.io/name: csi-secrets-store-provider-aws
+  podMetricsEndpoints:
+    - path: /metrics
+      port: metrics
+{{- end }}
--- a/modules/kubernetes/csi-secrets-store-provider-aws/charts/csi-secrets-store-provider-aws/Chart.yaml
+++ b/modules/kubernetes/csi-secrets-store-provider-aws/charts/csi-secrets-store-provider-aws/Chart.yaml
@@ -15,7 +15,7 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.1.1
+version: 0.1.0
 
 # This is the version number of the application being deployed. This version number should be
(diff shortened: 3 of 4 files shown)


CalleB3:
Please review and let me know if something needs changing :)

modules/kubernetes/aks-core/modules.tf
```diff
@@ -402,6 +402,7 @@ module "prometheus" {
   csi_secrets_store_provider_azure_enabled = var.csi_secrets_store_provider_azure_enabled
   aad_pod_identity_enabled                 = var.aad_pod_identity_enabled
   azad_kube_proxy_enabled                  = var.azad_kube_proxy_enabled
+  secrets_store_csi_driver_azure_enabled   = var.secrets_store_csi_driver_azure_enabled
```
bittrance, line 405: The naming (and possibly sorting/grouping) here is a bit confusing. How does `secrets_store_csi_driver_azure_enabled` and `csi_secrets_store_provider_azure_enabled` relate? It does not look as if they adhere to the same naming standard.

validation/kubernetes/eks-core/main.tf
```diff
@@ -88,7 +88,8 @@ module "eks_core" {
     volume_claim_storage_class_name = "default"
     volume_claim_size               = "5Gi"
 
-    alertmanager_enabled = false
+    alertmanager_enabled                 = false
+    secrets_store_csi_driver_aws_enabled = true
```
ahockersten, line 92: I don't understand why this is set to `true` here, but the default in `eks-core/variables.tf` is false. I expect them to be the same but maybe I'm missing something.

Same comment applies for `secrets_store_csi_driver_azure_enabled` above.

bittrance (changes requested):
I think the config variable needs a better name, but otherwise this looks kosher.

Having said that, I wonder why we do not always have a pod monitor in all namespaces, which looks for a generic label? Is there a reason we have so specific PodMonitor instances?

validation/kubernetes/eks-core/main.tf
NissesSenap: This is just validation config @ahockersten since it have a default value we don't even have to define it.

NissesSenap (commented):
As @bittrance and @phillebaba said we already have pointed out we already have: csi_secrets_store_provider_azure_enabled, we can use the same variable since these two functions are very tightly coupled. 
You can use csi_secrets_store_provider_azure_enabled, so all you need to do is to add 

So you more or less only need to add the PodMonitor in modules/kubernetes/prometheus/charts/prometheus-extras/templates/monitor.yaml
Don't forget to add enter at the end of your files.

NissesSenap (approved):
LGTM

NissesSenap:
@bittrance apparently I can't remove a change for request from you. Didn't know that github worked that way.
Please review again and we will be able to merge.

CHANGELOG.md
```diff
@@ -43,6 +48,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 - [#451](https://github.com/XenitAB/terraform-modules/pull/451) Set revision history for all certificates to limit the amount of certificate requests.
 
+### Changed
+
+- [#454](https://github.com/XenitAB/terraform-modules/pull/454) Set prometheus remote write queue config,
+  lowering default max shards and increasing default min back off.
+
```
NissesSenap, line 55: This shouldn't be changed, it have been fixed in another PR.

modules/kubernetes/prometheus/charts/prometheus-extras/templates/monitor.yaml
```diff
@@ -312,3 +312,37 @@ spec:
...
+metadata:
+  name: csi-secrets-store-driver-azure
+  namespace: csi-secrets-store-provider-azure
+  labels:
+    xkf.xenit.io/monitoring: platform
+spec:
+  selector:
+    matchLabels:
+      app.kubernetes.io/instance: csi-secrets-store-driver-azure
+      app.kubernetes.io/name: csi-secrets-store-driver-azure
```
NissesSenap, line 327: The value should be: csi-secrets-store-provider-azure

modules/kubernetes/prometheus/charts/prometheus-extras/templates/monitor.yaml
```diff
@@ -312,3 +312,37 @@ spec:
...
+metadata:
+  name: csi-secrets-store-driver-aws
+  namespace: csi-secrets-store-provider-aws
+  labels:
+    xkf.xenit.io/monitoring: platform
+spec:
+  selector:
+    matchLabels:
+      app.kubernetes.io/instance: csi-secrets-store-driver-aws
+      app.kubernetes.io/name: csi-secrets-store-driver-aws
```
NissesSenap, line 344: Today I don't define these values in the ds:
https://github.com/XenitAB/terraform-modules/blob/main/modules/kubernetes/csi-secrets-store-provider-aws/charts/csi-secrets-store-provider-aws/templates/daemonset.yaml

Please update the helm chart and don't forget to update: https://github.com/XenitAB/terraform-modules/blob/main/modules/kubernetes/csi-secrets-store-provider-aws/charts/csi-secrets-store-provider-aws/Chart.yaml

Set the values to:  csi-secrets-store-provider-aws to be consistent with azure

NissesSenap (changes requested):
Sorry should have thought about this earlier but now I have checked how it looks in reality. Please update the selectors to match.

phillebaba (approved):
LGTM

PR #440, opened 2021-11-05
Allow scale down for nodes with local storage

phillebaba:
This fix allows the autoscaler in EKS and AKS to scale down nodes with local storage. Without this the autoscaler will rarely scale down nodes. Users should not expect local storage to always be available so this should not cause any issues.

### DIFF
--- a/modules/azure/aks/aks.tf
+++ b/modules/azure/aks/aks.tf
@@ -9,15 +9,8 @@ resource "azurerm_kubernetes_cluster" "this" {
   api_server_authorized_ip_ranges = var.aks_authorized_ips
 
   auto_scaler_profile {
-    balance_similar_node_groups      = false
-    max_graceful_termination_sec     = 600
-    scale_down_delay_after_add       = "10m"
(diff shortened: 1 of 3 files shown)


modules/azure/aks/aks.tf
```diff
@@ -9,15 +9,8 @@ resource "azurerm_kubernetes_cluster" "this" {
...
   auto_scaler_profile {
-    balance_similar_node_groups      = false
-    max_graceful_termination_sec     = 600
-    scale_down_delay_after_add       = "10m"
-    scale_down_delay_after_delete    = "10s"
-    scale_down_delay_after_failure   = "3m"
-    scan_interval                    = "10s"
-    scale_down_unneeded              = "10m"
-    scale_down_unready               = "20m"
-    scale_down_utilization_threshold = "0.5"
```
phillebaba, line 20: No need to redefine default values

NissesSenap (approved):
Ahh that is really good, great news!

PR #441, opened 2021-11-07
Fix namespace resources dependencies

phillebaba:
This change fixes issues when creating new tenant namespaces as it would not wait for the namespace to be created before creating resources in it.

### DIFF
--- a/modules/kubernetes/eks-core/namespace.tf
+++ b/modules/kubernetes/eks-core/namespace.tf
@@ -7,6 +7,15 @@ resource "kubernetes_namespace" "service_accounts" {
   }
 }
 
+resource "kubernetes_service_account" "tenant" {
+  for_each = { for ns in var.namespaces : ns.name => ns }
+
+  metadata {
+    name      = each.value.name
+    namespace = kubernetes_namespace.service_accounts.metadata[0].name
+  }
+}
+
 resource "kubernetes_namespace" "tenant" {
   for_each = { for ns in var.namespaces : ns.name => ns }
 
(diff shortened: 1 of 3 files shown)


modules/kubernetes/eks-core/namespace.tf
```diff
@@ -25,36 +34,27 @@ resource "kubernetes_namespace" "tenant" {
...
-#      default_request = {
-#        cpu    = var.kubernetes_default_limit_range.default_request.cpu
-#        memory = var.kubernetes_default_limit_range.default_request.memory
-#      }
-#      default = {
-#        memory = var.kubernetes_default_limit_range.default.memory
-#      }
-#    }
-#  }
-#}
```
phillebaba, line 57: Not really sure why this was commented out. @NissesSenap  do you have any insight into this?

simongottschlag:
@phillebaba should the limit range fix be a separate PR? If it has been disabled for a while, it may be a breaking change and should be reflected as such in its own PR.

modules/kubernetes/eks-core/namespace.tf
NissesSenap: Don't know, my guess is that I wanted to keep everything as simple as possible initially and then I probably forgot to add it again.

phillebaba:
I have no idea why it was disabled in EKS as it has always been enabled in AKS. If there is a good reason for disabling it then sure lets do another PR, but if this is just a mistake I think its fine to do it in this PR.

NissesSenap (approved):

PR #442, opened 2021-11-08
ingress-nginx fixes

simongottschlag:
This PR enables tracing for datadog as well as adds `x-forwarded-for` header to services.

### DIFF
--- a/modules/kubernetes/ingress-nginx/templates/values.yaml.tpl
+++ b/modules/kubernetes/ingress-nginx/templates/values.yaml.tpl
@@ -60,7 +60,7 @@ controller:
     %{~ if datadog_enabled ~}
     ad.datadoghq.com/controller.check_names: '["nginx_ingress_controller"]'
     ad.datadoghq.com/controller.init_configs: '[{}]'
-    ad.datadoghq.com/controller.instances: '[{"prometheus_url": "http://%%host%%:10254/metrics"}]'
+    ad.datadoghq.com/controller.instances: '[{"prometheus_url": "http://%%host%%:%%port_metrics%%/metrics"}]'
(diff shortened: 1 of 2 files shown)


modules/kubernetes/ingress-nginx/templates/values.yaml.tpl
```diff
@@ -35,6 +35,11 @@ controller:
     ${key}: "${value}"
     %{~ endfor ~}
     server-tokens: "false"
+    use-forwarded-headers: "true"
```
NissesSenap, line 38: Is this something that we always wants to do?
Could any of our users be relying of this value being "false"?

modules/kubernetes/ingress-nginx/templates/values.yaml.tpl
```diff
@@ -65,6 +70,15 @@ controller:
     %{~ endif ~}
   %{~ endif ~}
 
+  %{~ if datadog_enabled ~}
```
NissesSenap, line 73: Why should this only be applied for datadog? Don't look like it would hurt the general config.

modules/kubernetes/ingress-nginx/templates/values.yaml.tpl
simongottschlag: We should always have X-Forwarded-For in my opinion.

modules/kubernetes/ingress-nginx/templates/values.yaml.tpl
simongottschlag: It's only being used by datadog as of now, which means it shouldn't be added to environments where it's not being used.

PR #451, opened 2021-11-16
Set revision history to all created cert manager certifiates

phillebaba:
This change sets a value for a Certificate revision history. This value controls how CertificateRequests are kept. The default value is nil, meaning CertificateRequests will never be cleaned up. Letsencrypt certificates dont really have this problem as they are only renewed every 3 months. However linkerd certificates will produce a lot of CertificateRequests as thos Certificates are renewed multiple times a day.

A logical reasoning would be that a reasonable default value would be set by the controller, but it seems like this will never happen.

Reference to the cert manager docs.
https://cert-manager.io/docs/reference/api-docs/#cert-manager.io/v1.Certificate

### DIFF
--- a/modules/kubernetes/linkerd/charts/linkerd-extras/templates/linkerd-identity-issuer.yaml
+++ b/modules/kubernetes/linkerd/charts/linkerd-extras/templates/linkerd-identity-issuer.yaml
@@ -4,6 +4,7 @@ metadata:
   name: linkerd-identity-issuer
 spec:
   secretName: linkerd-identity-issuer
+  revisionHistoryLimit: 3
   duration: 8h
   renewBefore: 4h
   issuerRef:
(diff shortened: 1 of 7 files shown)


NissesSenap (approved):
Nice find.
LGTM

modules/kubernetes/linkerd/charts/linkerd-extras/templates/linkerd-identity-issuer.yaml
```diff
@@ -4,6 +4,7 @@ metadata:
   name: linkerd-identity-issuer
 spec:
   secretName: linkerd-identity-issuer
+  revisionHistoryLimit: 3
```
simongottschlag, line 7: Perhaps we should have the linkerd certs a bit higher? Saving it for 4 days at least? (Something happens on Friday, we look at it on Monday)

PR #454, opened 2021-11-17
Prometheus set queue config in remote write

NissesSenap:
This to be nicer to thanos receiver when sending data to it.
We can wait a few extra miliseconds for this data to come to the receiver, as long as it comes.


phillebaba (approved):
LGTM

PR #456, opened 2021-11-17
Remove goldpinger from aks and eks core and deprecated goldping module

phillebaba:
I have for a while been seeing a lot of false positives from goldpinger that just dont make sense. Personally I do not see a need for goldpinger as it does not do the job it advertises well. I think it is best to just remove goldping as nobody relies on it as a health metrics. One less daemonset in our clusters.


phillebaba:
Will fix the changelog later after discussions.

simongottschlag (dismissed):
Looking good! Can we move it to deprecated modules in the readme at the same time?

NissesSenap (approved):

PR #457, opened 2021-11-17
Fix prometheus remote write queue config

phillebaba:
This changes fixes some wrong conclusions we made previously in regard to fixes in the remote write. This change is based on a change to the default value in the next release of Prometheus.

https://github.com/prometheus/prometheus/pull/9634


NissesSenap (approved):
LGTM

PR #459, opened 2021-11-18
Decrease max shards

phillebaba:
This change decreases the max shards from the default 200 to 100. The reasoning for this is to reduce the risk of sending a large amount of requests when remote write starts to fail. By reducing the max shard count we reduce the concurrency for each Prometheus instance.


NissesSenap (approved):
LGTM

PR #465, opened 2021-11-25
Move xenit credentials to Prometheus and remove Xenit proxy

phillebaba:
This change will move the tls credentials directly to Prometheus and removes the Xenit proxy.

The reasoning behind this is the issues related to SNAT port exhaustion that can be caused by Prometheus remote write. These issues are not apparent immediately as long as all requests succeed. Issues happen when the receiver stops working and Prometheus attempts to catch up with the cached metrics. In these cases the amount of requests will increase. Removing the proxy will decrease the SNAT usage, mitigating these issues. The reasons for this is currently unknown.

For more details about these issues refer to the following page. 
https://www.danielstechblog.io/detecting-snat-port-exhaustion-on-azure-kubernetes-service/


modules/kubernetes/prometheus/charts/prometheus-extras/templates/prometheus.yaml
```diff
@@ -1,11 +1,13 @@
 apiVersion: monitoring.coreos.com/v1
 kind: Prometheus
 metadata:
+  name: prometheus-xks
```
NissesSenap, line 4: Should we change the name to `xks` instead? 
Right now the generated pods are called: `prometheus-prometheus-xks-0`

modules/kubernetes/prometheus/charts/prometheus-extras/templates/prometheus.yaml
```diff
@@ -1,11 +1,13 @@
...
   labels:
     {{- include "prometheus-extras.labels" . | nindent 4 }}
-  name: prometheus-xks
-  namespace: prometheus
 spec:
+  podMetadata:
+    labels:
+      aadpodidbinding: prometheus
```
NissesSenap, line 10: Should be set within a if for cloud = azure

NissesSenap (commented):
Minor comments, great job.

NissesSenap (approved):

PR #467, opened 2021-11-29
Change the method of checking changelog updates

phillebaba:
This change fixes the changelog check for PRs originating from forks. The new checker will also check the content of the changelog to make sure that it is correct.


bittrance:
PR description, plz. Apart from answering why, it might be worthwhile mentioning that we switch from `ignore-changelog` -> `skip-changelog`. Are there any other differences?

bittrance (approved):
Awesome: FOSS FTW.

PR #470, opened 2021-11-30
Set max history for helm releases

phillebaba:
This change will reduce the amount of secrets created by Helm to store chart history. The default max history is 0 meaning that there is no limit. We dont really use the chart history as we do not rollback on failure. Setting it to 3 is just what I think is a reasonable number.


NissesSenap (dismissed):
3 sounds like a reasonable nr.

phillebaba:
@NissesSenap need another review after rebase.

NissesSenap (dismissed):

NissesSenap (approved):
LGTM

PR #472, opened 2021-12-06
Ingress block annotation

NissesSenap:
Update ingress-nginx to 3.40.0  and disable allow-snippet-annotations by default. Add [annotation-value-word-blocklist](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#annotation-value-word-blocklist)
to mitigate CVE-2021-25742


NissesSenap:
Played a bit with the settings and if you try to add a server-snippet with _by_lua

You will see the following as an error:

```
admission webhook "validate.nginx.ingress.kubernetes.io" denied the request: nginx.ingress.kubernetes.io/server-snippet annotation contains invalid word _by_lua
```

NissesSenap:
Or if you disable annotations completely you will get:

```
admission webhook "validate.nginx.ingress.kubernetes.io" denied the request: nginx.ingress.kubernetes.io/server-snippet annotation cannot be used. Snippet directives are disabled by the Ingress administrator
```

Funny thing though, it seems like the annotation block took it before the disable block. Since I have both settings you will see the first error.
It was only after i deleted the annotation-value-word-blocklist that I was able to see this error.

phillebaba:
I think this looks good but I want @simongottschlag input also.

modules/kubernetes/ingress-nginx/templates/values.yaml.tpl
```diff
@@ -44,6 +44,8 @@ controller:
     http-snippet: |
       ${http_snippet}
     %{~ endif ~}
+    allow-snippet-annotations: ${allow_snippet_annotations_enabled}
+    annotation-value-word-blocklist: load_module,lua_package,_by_lua,location,root,proxy_pass,serviceaccount,{,},',\
```
simongottschlag, line 48: This seems dangerous. Is it tested?

simongottschlag (approved):
LGTM. Left a comment in regards to the blocklist

modules/kubernetes/ingress-nginx/templates/values.yaml.tpl
NissesSenap: Yes, look at comments earlier in the PR and you see how the config respons to the same config defined in the hackerone example.

NissesSenap: On the other hand I'm a bit confused on how it works. The disable flag only seems to be on custom server annotation and not things like the whitelist annotation. I can read some more code and be sure that i have understood when it applies and not. I do think this is what we want but it's never bad to understand more :)

NissesSenap:
So i read the ingress-nginx controller code and found this:

https://github.com/kubernetes/ingress-nginx/blob/5cff197bc51079f18a9714c69a0dd11e4bcf310b/internal/ingress/controller/controller.go#L267-L269

It's only if the annotation have `-snippet` the rule AllowSnippetAnnotations will give deny. So for example if you are using https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#http-snippet you have to set the config value: `allow_snippet_annotations_enabled = true` or this new PR will break your ingress config.

Knowing this I think we should be okay to merge this PR. I hope you agree @simongottschlag and @phillebaba

modules/kubernetes/eks-core/modules.tf
```diff
@@ -140,11 +140,12 @@ module "ingress_nginx" {
...
-  http_snippet           = var.ingress_config.http_snippet
-  linkerd_enabled        = var.linkerd_enabled
-  datadog_enabled        = var.datadog_enabled
-  public_private_enabled = var.ingress_config.public_private_enabled
+  cloud_provider                    = "aws"
+  http_snippet                      = var.ingress_config.http_snippet
+  linkerd_enabled                   = var.linkerd_enabled
+  datadog_enabled                   = var.datadog_enabled
+  public_private_enabled            = var.ingress_config.public_private_enabled
+  allow_snippet_annotations_enabled = var.ingress_config.allow_snippet_annotations_enabled
```
simongottschlag, line 148: Please change to either `allow_snipppet_annotations` or `snippet_annotations_enabled`.

NissesSenap: allow_snipppet_annotations it is

CHANGELOG.md
```diff
@@ -6,19 +6,24 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
 
+### Added
+
+- [#463](https://github.com/XenitAB/terraform-modules/pull/463) Add azure-metrics to monitor azure specific metrics.
+
 ### Changed
 
 - [#470](https://github.com/XenitAB/terraform-modules/pull/470) Set max history for Helm releases to reduce the amount of secrets created.
 - [#471](https://github.com/XenitAB/terraform-modules/pull/471) Update azad-kube-proxy from v0.0.27 to v0.0.30 and remove dashboard (k8dash/skooner)
+- [#472](https://github.com/XenitAB/terraform-modules/pull/472) [Breaking] Update ingress-nginx to 3.40.0
```
phillebaba, line 17: Shorten, keep details in PR description.

phillebaba (approved):

NissesSenap:
From the old changelog.

[Breaking] Update ingress-nginx to 3.40.0   and disable allow-snippet-annotations by default. Add [annotation-value-word-blocklist](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#annotation-value-word-blocklist).
to mitigate [CVE-2021-25742](https://github.com/kubernetes/ingress-nginx/issues/7837).
The config allow_snippet_annotations only applies to annotations that contain -snippet,
for [example http-snippet](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/configmap/#http-snippet).

PR #473, opened 2021-12-08
Gather trivy metrics using starboard-exporter

NissesSenap:
Currently starboard don't expose any metrics from trivy CVE reports.
There is a discussion ongoing in the starboard community about this but it will take some time before anything happens.
In the meantime let's use starboard-exporter.


NissesSenap:
There is a helm chart in: https://github.com/giantswarm/starboard-exporter and it can be installed using:

```shell
helm repo add giantswarm https://giantswarm.github.io/giantswarm-catalog
helm repo update
helm upgrade -i --name starboard-exporter --namespace starboard-operator --version 0.1.2 giantswarm/starboard-exporter
``` 

But currently it got a bit to much rbac access and enforces PSP so that is why I use my own chart in this PR. I have created a PR: https://github.com/giantswarm/starboard-exporter/pull/27 with them where I have done some cleanup but more is needed.

When this PR get's merged I will create a new issue going over to there helm chart instead

NissesSenap:
Starboard is apparently not available in eks core. I will create a seperate PR to fix this. I still add the variable so we can add and disable the metric to be consistent with aks core.

NissesSenap:
The metric that gets generated looks like this:
```
starboard_exporter_vulnerabilityreport_image_vulnerability_severity_count{container="starboard-exporter", endpoint="metrics", image_digest="sha256:e9fb216ace49dfa4a5983b183067e97496e7a8b307d2093f4278cd550c303899", image_namespace="ingress-nginx", image_repository="ingress-nginx/controller", instance="10.244.2.126:8080", job="starboard-exporter", namespace="starboard-operator", pod="starboard-exporter-76c8f9c44b-lcw6m", report_name="replicaset-ingress-nginx-controller-5f5977bd4-controller", service="starboard-exporter", severity="HIGH"}
```

The starboard-exporter team have decided to minimize the number of labels that generated by only provide a count of how many CVE:s and it's different severity in there metric by default. This is possible to change if we want more information. 
But personally I agree with them and think this should be a indicator about something that you need to look deeper in to.

But if want to get the exact CVE id that is possible but it will generate lots of metrics.

phillebaba (approved):

Issue #303, opened 2021-06-07
Only send next_hop_in_ip_address when next_hop_type is VirtualAppliance


jimmydao:
Azure only allow next_hop_in_ip_address when next_hop_type is VirtualAppliance. 
Is it possible to only send next_hop_in_ip_address when when type is VirtualAppliance? 

https://github.com/XenitAB/terraform-modules/blob/307192c7d7729f273421c281747aab276c6626b2/modules/azure/core/routes.tf#L23

PR #482, opened 2021-12-14
azure/core: add support for non VirtualAppliance routes

simongottschlag:
This fixes the issue described in the following issue:
https://github.com/XenitAB/terraform-modules/issues/303

Now we're forking on the next_hop_type and remove the
next_hop_in_ip_address if it's not a VirtualAppliance route.

Let's hope it solves the problem. :^)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## Unreleased
 
+### Changed
+
+- [#482](https://github.com/XenitAB/terraform-modules/pull/482) azure/core: add support for non VirtualAppliance routes
+
 ## 2021.12.2
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## Unreleased
 
+### Changed
+
+- [#482](https://github.com/XenitAB/terraform-modules/pull/482) azure/core: add support for non VirtualAppliance routes
```
phillebaba, line 11: No prefix needed

modules/azure/core/routes.tf
```diff
@@ -23,6 +24,20 @@ resource "azurerm_route" "this" {
   next_hop_in_ip_address = each.value.route.next_hop_in_ip_address
 }
 
+resource "azurerm_route" "not_virtual_appliance" {
```
phillebaba, line 27: Please add a note about how we should add move commands in the future when we upgrade to Terraform 1.1.0.
https://github.com/hashicorp/terraform/releases

NissesSenap (approved):
LGTM what a dream team <3

PR #484, opened 2021-12-14
Enable node encryption for all AKS nodes

phillebaba:
This change will force all nodes to enable encryption at rest. Two current issues with enabling this. One is that it will require node recreation, the other is that it seems like the feature needs to be enabled per subscription.

More info can be found here.
https://docs.microsoft.com/en-us/azure/aks/enable-host-encryption


phillebaba:
@simongottschlag @NissesSenap @jimgus 

Considering to just close this because we have moved over to ephemeral nodes. We will never have any data at rest right?

NissesSenap:
Sounds reasonable

simongottschlag:
@phillebaba data will be stored on the nodes the VMs are running and I think it would we good to have the facts about how it's handled before closing.

landerss1 (changes requested):
Changing the value of enable_host_encryption requires specifying [temporary_name_for_rotation](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster#temporary_name_for_rotation).

PR #486, opened 2021-12-14
Add configuration options for spot instances

phillebaba:
This change makes it possible to add spot instances to an AKS cluster.


NissesSenap (approved):
LGTM, had to learn a bit about validation as well. Haven't used it before :)

PR #491, opened 2021-12-21
kubernetes/grafana-agent: new module for observability with Grafana Cloud

simongottschlag:
This PR introduces the Grafana Agent which will initially be used to send metrics, logs and traces to Grafana Cloud.


modules/kubernetes/grafana-agent/charts/grafana-agent-extras/values.yaml
```diff
@@ -0,0 +1,10 @@
+image: grafana/agent:v0.21.2
+replicas: 2
+environment: dev
+clusterName: aks1
+credentialsSecretName: grafana-agent-credentials
+
+remote:
+  metricsUrl: https://prometheus-prod-01-eu-west-0.grafana.net/api/prom/push
+  logsUrl: https://logs-prod-eu-west-0.grafana.net/loki/api/v1/push
+  tracesUrl: tempo-eu-west-0.grafana.net:443
```
phillebaba, line 10: NIT: Should we have default values for EU or just set empty strings instead?

simongottschlag: fixed!

phillebaba (approved):

PR #497, opened 2021-12-27
Remove namespace config for kube_state_metrics

jimgus:
The namespaces config option was introduced to be able to limit the metrics volume from kube-state-metrics. This was introduced by #448 but it also introduced a bug that the same cluster role was created multiple times and hence it was not possible to deploy the prometheus module. This change reverts the above PR since we found that this feature was of very limited value anyway.


simongottschlag:
@jimgus please add a description to the PR and describe why this is needed

simongottschlag (commented):
LGTM :)

simongottschlag (approved):

jimgus:
The namespaces config option was introduced to be able to limit the metrics volume from kube-state-metrics. This was introduced by https://github.com/XenitAB/terraform-modules/pull/448 but it also introduced a bug that the same cluster role was created multiple times and hence it was not possible to deploy the prometheus module. This change reverts the above PR since we found that this feature was of very limited value anyway.

PR #498, opened 2021-12-27
Set AKS cluster autoscaler expander strategy to least waste

phillebaba:
This changes the expander strategy from the default random to least waste. Hopefully this will result in the cluster autoscaler selecting the right type of instance class when scaling up. It is possible to select multiple expanders but that is not currently supported by AKS. There is an issue in the provider which tracks this hashicorp/terraform-provider-azurerm/issues/13900.

Refer to the cluster autoscaler documentation for detailed information.
https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/FAQ.md#what-are-expanders

### DIFF
--- a/modules/azure/aks/aks.tf
+++ b/modules/azure/aks/aks.tf
@@ -17,6 +17,9 @@ resource "azurerm_kubernetes_cluster" "this" {
   auto_scaler_profile {
     # Pods should not depend on local storage like EmptyDir or HostPath
     skip_nodes_with_local_storage = false
+    # Selects the node pool which would result in the least amount of waste.
+    # TODO: When supported we should make use of multiple expanders #499
+    expander = "least-waste"
   }
 
   network_profile {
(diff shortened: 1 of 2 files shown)


phillebaba:
In the future we should consider how we should prioritize spot instances and then fallback to on demand if possible. But right now I do not think it should be prio.

modules/azure/aks/aks.tf
```diff
@@ -17,6 +17,9 @@ resource "azurerm_kubernetes_cluster" "this" {
   auto_scaler_profile {
     # Pods should not depend on local storage like EmptyDir or HostPath
     skip_nodes_with_local_storage = false
+    # Selects the node pool which would result in the least amount of waste.
+    # TODO: When supported we should make use of multiple expanders
```
simongottschlag, line 21: Please create issue in this repo about this and link to that issue here.

simongottschlag (dismissed):
One small comment, otherwise looks good! :)

simongottschlag (approved):

Issue #495, opened 2021-12-23
grafana-agent: cluster shows up as grafana-agent/grafana-agent for all environments


simongottschlag:
We need a way to separate logs in different environments, right now they all show up as cluster `grafana-agent/grafana-agent`.

Related issue: https://github.com/grafana/agent/issues/1035

PR #502, opened 2021-12-28
grafana-agent: Add externalLabels to logs

simongottschlag:
Please note that `externalLabels` are set in two different places for logs and metrics, which is beautifully described in the following comment: https://github.com/grafana/agent/issues/1035#issuecomment-953607160

We are also changing the name from `clusterName` to `cluster` which replaces the "default" cluster label (which was `grafana-agent/grafana-agent`). This may, or may not, cause issues in the future - but I've been promised by @NissesSenap that he'll solve it using jsonnet. ;^)

This will resolve the following issue: https://github.com/XenitAB/terraform-modules/issues/495

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## Unreleased
 
+### Changed
+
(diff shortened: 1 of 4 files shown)


NissesSenap (approved):
LGTM

NissesSenap (approved):

PR #504, opened 2021-12-29
Give users access to read/delete starboard reports

NissesSenap:
In the future developers should be able to use the reports generated from starboard to gain insights on how there configuration of their deployments is done together with CVE:s.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## Unreleased
 
+### Added
+
+- [#504](https://github.com/XenitAB/terraform-modules/pull/504) Give developers access to starboards report.
+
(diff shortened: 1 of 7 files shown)


NissesSenap:
To tiered to find the issue:

│ Error: Invalid index
│ 
│   on .terraform/modules/aks1_core/modules/kubernetes/aks-core/k8s-role-binding.tf line 183, in resource "kubernetes_role_binding" "starboard_reports":
│  183:     name      = kubernetes_cluster_role.starboard_reports[each.key].metadata[0].name
│     ├────────────────
│     │ each.key is "lab"
│     │ kubernetes_cluster_role.starboard_reports is object with 1 attribute "starboard"
│ 
│ The given key does not identify an element in this collection value.

NissesSenap:
Think what a clear mind can do, solved in 5 minutes :)

jimgus (approved):
LGTM!

Issue #505, opened 2021-12-30
Feature enable VPA(Vertical Pod Autoscaling) recommendations in XKF


NissesSenap:
For us to be able to give simple suggestions to our customers we should install VPA in our clusters and enable it in recommendation mode. This way we can easily provide an overview of this data in our reports and it will also be simple to use for our developers.

One potential solution could be to use FairwindowsOps helm chart since VPA don't supply a official helm chart: https://github.com/FairwindsOps/charts/tree/master/stable/vpa
https://goldilocks.docs.fairwinds.com/installation/#installation-2

To get VPA to work we would need to enable it in kube state metrics: https://github.com/kubernetes/kube-state-metrics/blob/master/docs/verticalpodautoscaler-metrics.md

Great overview of VPA: https://povilasv.me/vertical-pod-autoscaling-the-definitive-guide/#

There are a two grafana dashboards that we can take inspiration from: https://grafana.com/grafana/dashboards/?search=vpa

NissesSenap:
Installed goldilocks manually https://goldilocks.docs.fairwinds.com/ in our dev clusters after applying VPA and helps out to generate the VPA config.

The VPA it generates looks like this
```shell
$ k get verticalpodautoscaler goldilocks-podinfo -o yaml
apiVersion: autoscaling.k8s.io/v1
kind: VerticalPodAutoscaler
metadata:
  labels:
    creator: Fairwinds
    source: goldilocks
  name: goldilocks-podinfo
  namespace: lab
spec:
  targetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: podinfo
  updatePolicy:
    updateMode: "Off"
Status:
  Conditions:
    Last Transition Time:  2021-12-30T14:18:59Z
    Status:                True
    Type:                  RecommendationProvided
  Recommendation:
    Container Recommendations:
      Container Name:  podinfo
      Lower Bound:
        Cpu:     15m
        Memory:  29865759
      Target:
        Cpu:     15m
        Memory:  36253748
      Uncapped Target:
        Cpu:     15m
        Memory:  36253748
      Upper Bound:
        Cpu:     1130m
        Memory:  3725681106
Events:          <none>
```

and if you use there dashboard you will see something like this:

![image](https://user-images.githubusercontent.com/14774899/147760119-31137f68-211a-48c9-a974-cc73f299da3d.png)

[...]

PR #506, opened 2021-12-30
Setup VPA and use goldilocks to autogenerate VPA requests

NissesSenap:
For us to be able to give simple suggestions to our customers we should install VPA in our clusters and enable it in recommendation mode. This way we can easily provide an overview of this data in our reports and it will also be simple to use for our developers.

Partly solves: https://github.com/XenitAB/terraform-modules/issues/505


modules/kubernetes/vpa/templates/vpa-values.yaml.tpl
```diff
@@ -0,0 +1,36 @@
+priorityClassName: platform-medium
+
+recommender:
+  # recommender.enabled -- If true, the vpa recommender component will be installed.
+  enabled: true
+  # recommender.extraArgs -- A set of key-value flags to be passed to the recommender
+  extraArgs:
+    v: "4"
```
NissesSenap, line 8: Should lower this logging, VPA generates lots of logs

modules/kubernetes/vpa/templates/vpa-values.yaml.tpl
```diff
@@ -0,0 +1,36 @@
...
+    pod-recommendation-min-cpu-millicores: 15
+    pod-recommendation-min-memory-mb: 24
+    prometheus-address: http://prometheus-operated.prometheus.svc.cluster.local:9090
+    storage: prometheus
+    # How much time back prometheus have to be queried to get historical metrics
+    # history-length: 8
```
NissesSenap, line 14: Got `Could not initialize history provider: history length 8 is not a valid Prometheus duration: not a valid duration string: 8`
Probably need to define if it should be hours days or something like that. Need to read the docs. 8 days is the default value

NissesSenap:
* Currently haven't added if in kube-state-metrics to add VPA.
* Haven't added this feature to EKS.
* Should add podmonitor config for prometheus of the VPA.

jimgus (commented):
I think it can be good to mention somewhere in the Readme that we only use the recommend part of VPA

NissesSenap:
The current plan is to use goldilocks to auto create VPA config.
There are two ways to go, ether we add two flags:

```
--on-by-default
--exclude-namespaces=demo-excluded
```

or we use the default mode and label all our namespaces with:
`goldilocks.fairwinds.com/enabled=true`

I think it's better that we use `on-by-default`.
I'm unsure if we think we need to disable any namespaces.

NissesSenap:
I have excluded goldilocks from creating VPA in kube-system since we don't control the resources in there.
I have added a extra rbac rule so we can create VPA rules for cronjobs.
Ready for review, @jimgus @phillebaba

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## Unreleased
 
+### Added
+
+- [#506](https://github.com/XenitAB/terraform-modules/pull/506) Add VPA(Vertical Pod Autoscaling) as a module.
```
phillebaba, line 11: NIT: Add a space after VPA

modules/kubernetes/prometheus/templates/values.yaml.tpl
```diff
@@ -31,11 +31,44 @@ kube-state-metrics:
...
+    - replicationcontrollers
+    - resourcequotas
+    - secrets
+    - services
+    - statefulsets
+    - storageclasses
+    - validatingwebhookconfigurations
+    - volumeattachments
+    - verticalpodautoscalers # not a default resource, see also: https://github.com/kubernetes/kube-state-metrics#enabling-verticalpodautoscalers
+  %{ endif }
```
phillebaba, line 65: Seems stupid that we have to redefine all collectors but there does not seem to be another way :(

modules/kubernetes/vpa/templates/vpa-values.yaml.tpl
```diff
@@ -0,0 +1,35 @@
...
+  image:
+    # recommender.image.repository -- The location of the recommender image
+    repository: k8s.gcr.io/autoscaling/vpa-recommender
+    # recommender.image.pullPolicy -- The pull policy for the recommender image. Recommend not changing this
+    pullPolicy: Always
```
phillebaba, line 18: Why are we setting pull policy Always? We should really avoid using this.

phillebaba (approved):

PR #514, opened 2022-01-14
Starboard scan for more severitys

NissesSenap:
* Upgrade trivy and starboard helm chart


NissesSenap:
I decided against adding lower severity errors. It's a risk that it will create to much data and etcd will say request is to large: https://github.com/aquasecurity/starboard/issues/899
Also if you fix the rest of your CVE:s you most likely fix the majority of the lower once as well.
And even if you don't, they are lower for a reason.

If someone want to have lower in the report in the future we can add it as a variable.

phillebaba (approved):

PR #517, opened 2022-01-17
Change VPA storage from prometheus to checkpoint

phillebaba:
Enabling agent mode in Prometheus removes the ability to run queries. The current VPA configuration is setup to use cadvisor metrics from Promtheus to calculate its recommendation. This is not possible anymore. The fix right now is to switch to the default checkpoint method which stores metrics in a CRD.

We should in the future have a look at if this is the best solution, or if there is a better one out there.

### DIFF
--- a/modules/kubernetes/vpa/templates/vpa-values.yaml.tpl
+++ b/modules/kubernetes/vpa/templates/vpa-values.yaml.tpl
@@ -1,23 +1,9 @@
 priorityClassName: platform-medium
 
 recommender:
-  # recommender.enabled -- If true, the vpa recommender component will be installed.
-  enabled: true
-  # recommender.extraArgs -- A set of key-value flags to be passed to the recommender
   extraArgs:
     pod-recommendation-min-cpu-millicores: 15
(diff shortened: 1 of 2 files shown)


modules/kubernetes/vpa/templates/vpa-values.yaml.tpl
```diff
@@ -1,23 +1,9 @@
 priorityClassName: platform-medium
 
 recommender:
-  # recommender.enabled -- If true, the vpa recommender component will be installed.
-  enabled: true
```
NissesSenap, line 5: Is enabled by default?

NissesSenap (approved):
LGTM

modules/kubernetes/vpa/templates/vpa-values.yaml.tpl
phillebaba: Si ;)
https://github.com/FairwindsOps/charts/blob/b74f826d24426b97d64081f8344495c0a33dfebd/stable/vpa/values.yaml#L29

PR #519, opened 2022-01-19
Fix nginx ingress service monitor selector when running multiple controllers

### LINKED ISSUE #518: Ingress Nginx Service Monitor when runing private/public
Currently the ServiceMonitor does not work for Ingress Nginx when running in private/public. This is due to the label selectors being incorrect . Fix should quick, the only consideration is if a single ServiceMonitor will work for two deployments without causing time series overlap.

phillebaba:
This change fixes the labels and label selector for the metrics services. Issues occur when private/public controllers are enabled, which cause the service monitor label selector to fail. This fix changes the labels so that both metrics services can be selected with the same selector.

Fixes #518


NissesSenap (approved):
LGTM

PR #520, opened 2022-01-19
Pull request template

bittrance:
This PR introduces a PR template built mostly on HTML comments. Most PR templates out there come with headings and example text. The benefit of using HTML comments is that you can just ignore the comments and type what you want. If you are in a hurry, you just start typing at the top and the result will be what you expect.

I'm not a frequent contributor to terraform-modules, so I'm not sure exactly what recommendations are needed. Please suggest any content you think should be included. I have chosen not to say anything about what is checked by status checks, since they will explain what is wrong when they fail.

@CalleB3 requested a PR template for this repo in order to help understand what should be in a PR description.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -24,5 +24,5 @@ Completion workflow:
 
 - [ ] Ensure all pull request comments are addressed
 - [ ] Rebase your branch against main branch
-- [ ] Ensure that all checks passes with the latest changes on main branch
+- [ ] Ensure that all checks pass with the latest changes on main branch
 - Merge the pull request.
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,9 +1,5 @@
 <!--
-    If you are changing an existing module or resource, describe why that configuration is no longer acceptable. If you are fixing a misconfiguration, make sure to describe both the intended result and the actual (faulty) result.
--->
-
-<!--
-    If you are adding a module or resources, describe what usage will be supported when this change is applied.
+    If you are adding a module or resources, describe what usage will be supported when this change is applied. If you are changing an existing module or resource, describe why that configuration is no longer acceptable. If you are fixing a misconfiguration, make sure to describe both the intended result and the actual (faulty) result.
 -->
 
 <!--
@@ -22,7 +18,8 @@ Fixes #NNN <!-- This will close the issue automatically when you merge -->
 
 Completion workflow:
 
-- [ ] Ensure all pull request comments are addressed
+- [ ] Module docs reflect these changes
+- [ ] All pull request comments are addressed
 - [ ] Rebase your branch against main branch
-- [ ] Ensure that all checks pass with the latest changes on main branch
+- [ ] All checks pass with the latest changes on main branch
 - Merge the pull request.
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,28 @@
+<!--
+    If you are changing an existing module or resource, describe why that configuration is no longer acceptable. If you are fixing a misconfiguration, make sure to describe both the intended result and the actual (faulty) result.
+-->
+
+<!--
+    If you are adding a module or resources, describe what usage will be supported when this change is applied.
+-->
+
+<!--
+    Explain what alternative configurations you have evaluated and why the chosen configuration is preferable.
+-->
+
+<!--
+    What do reviewers need to know to review? Link pull requests or issues that are related.
+-->
+
+<!--
+    What do we need to consider when rolling out this change?
+-->
+
+Fixes #NNN <!-- This will close the issue automatically when you merge -->
+
+Completion workflow:
+
+- [ ] Ensure all pull request comments are addressed
+- [ ] Rebase your branch against main branch
+- [ ] Ensure that all checks passes with the latest changes on main branch
+- Merge the pull request.


CalleB3 (dismissed):
Looks good 👍🏼

CalleB3:
Looks good! Maybe we can have someting similar for other repos aswell?

.github/pull_request_template.md
```diff
@@ -0,0 +1,28 @@
...
+    What do we need to consider when rolling out this change?
+-->
+
+Fixes #NNN <!-- This will close the issue automatically when you merge -->
+
+Completion workflow:
+
+- [ ] Ensure all pull request comments are addressed
+- [ ] Rebase your branch against main branch
+- [ ] Ensure that all checks passes with the latest changes on main branch
```
ahockersten, line 27:
```suggestion
- [ ] Ensure that all checks pass with the latest changes on main branch
```
Isn't this implicitly guarded against by not allowing merges when CI pipelines fail? In other words it can probably be removed?

.github/pull_request_template.md
```diff
@@ -0,0 +1,28 @@
+<!--
+    If you are changing an existing module or resource, describe why that configuration is no longer acceptable. If you are fixing a misconfiguration, make sure to describe both the intended result and the actual (faulty) result.
```
ahockersten, line 2: Maybe suggest having a test for the original failure, that now passes with the change?

bittrance: This is IaC so there are typically no tests.

ahockersten:
Love the usage of HTML comments in this way, it's my preferred way of doing things as well.

CalleB3 (approved):

ahockersten (dismissed):

NissesSenap (commented):
I like the comments but I would add a  headers for the different comment to make sure that your write something and you will get a decent structure out of the box.

.github/pull_request_template.md
```diff
@@ -0,0 +1,28 @@
...
+-->
+
+<!--
+    If you are adding a module or resources, describe what usage will be supported when this change is applied.
```
jimgus, line 6: I think this information belong to the documentation so I don't think it is needed to be repeated in the PR description.

.github/pull_request_template.md
```diff
@@ -0,0 +1,28 @@
...
+<!--
+    Explain what alternative configurations you have evaluated and why the chosen configuration is preferable.
+-->
+
+<!--
+    What do reviewers need to know to review? Link pull requests or issues that are related.
+-->
+
+<!--
+    What do we need to consider when rolling out this change?
```
jimgus, line 18: I kind of like the approach of providing that information in the change log instead of having tho browse all PRs, e.g.  https://github.com/XenitAB/terraform-modules/releases/tag/2022.01.2

.github/pull_request_template.md
```diff
@@ -0,0 +1,28 @@
...
+-->
+
+Fixes #NNN <!-- This will close the issue automatically when you merge -->
+
+Completion workflow:
+
+- [ ] Ensure all pull request comments are addressed
+- [ ] Rebase your branch against main branch
+- [ ] Ensure that all checks pass with the latest changes on main branch
+- Merge the pull request.
```
jimgus, line 28: Different way of doing things but I like the changes in a PR to be squashed unless there are commits that need to be in the history separately (not that common)

.github/pull_request_template.md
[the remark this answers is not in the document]
jimgus: Maybe it would be possible to generalize this paragraph to explain why this change is needed?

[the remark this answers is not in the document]
bittrance: https://www.terraform.io "Terraform codifies cloud APIs into declarative configuration files." Their docs consistently call the content of `.tf` files "configuration".

.github/pull_request_template.md
bittrance: The top comment is only about changes and fixes so you write for either of these, but typically not both. I'll see if I can rearrange the text to make this clearer.

.github/pull_request_template.md
bittrance: You're right. I'll change it to a completeness checklist with docs, release notes, et c; the things the checks don't cover.

.github/pull_request_template.md
bittrance: I think I do not understand the comment. Is there anything in the list that should be different? I'm deliberately not saying anything about how you craft the commits because it varies with taste and task at hand, so long as you use merge rather than fast-forward.

.github/pull_request_template.md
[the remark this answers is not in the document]
jimgus: I see, perfectly fine then. My bad

.github/pull_request_template.md
jimgus: Ok, I did the wrong interpretation here. I was thinking of the different PR options “Merge” and “Squash and merge”

.github/pull_request_template.md
bittrance: @NissesSenap is of the opinion that we should not include details about breaking changes in the changelog itself. If that's the case, then it will have to be in the PR description, whence they can be copied to release notes. I don't have a strong opinion either way.

.github/pull_request_template.md
bittrance: Merged the add and change section into one. See if this read better?

.github/pull_request_template.md
```diff
@@ -0,0 +1,28 @@
...
+-->
+
+Fixes #NNN <!-- This will close the issue automatically when you merge -->
+
+Completion workflow:
+
+- [ ] Module docs reflect these changes
+- [ ] All pull request comments are addressed
+- [ ] Rebase your branch against main branch
+- [ ] All checks pass with the latest changes on main branch
```
phillebaba, line 27: Some of these things are already automatic checks. Others should probably be resolved through automatic checks.

phillebaba:
My personal opinion is that parts of the template are too detailed. This template will only be applicable to specific PRs with new feature implementations but will not be relevant for smaller changes. My experience with other projects is that it becomes cumbersome when the PR template requires a lot more information than is actually necessary.

.github/pull_request_template.md
bittrance: I tried to avoid those that are automatic except the final pass check which is more motivated by the workflow; I suppose it could be folded into the rebase entry. What other items are covered by automatic checks? (We could of course turn on "Require conversation resolution before merging".)

bittrance:
> requires a lot more information than is actually necessary

@phillebaba  The template does not intend to require any information. It intends to help you to consider what is pertinent for your pull request. How can we change the text to make this clearer?

bittrance:
I do not really have time to work with this now and it seems the need has subsided for the time being.

PR #522, opened 2022-01-20
Add networkpolicy datadog and grafana-agent in tenant namespace

NissesSenap:
This to make it easier for our developers.
If they want a specific monitoring solution they should have networkpolicyies
inplace to do so straight away.


NissesSenap:
Have verified that both networkpolicies works as intended, this PR is ready for review.

modules/kubernetes/eks-core/variables.tf
```diff
@@ -340,3 +340,9 @@ variable "vpa_enabled" {
   type        = bool
   default     = true
 }
+
+variable "grafana_agent_enabled" {
```
NissesSenap, line 344: Should we do this or should we skip grafana_agent_enabled in eks since we haven't verified it yet and we probably need to do a few things. My thought was that I wanted to keep the networkpolicies as similar as possible as we talked about on yesterdays meeting.

jimgus: Keep it!

jimgus (approved):
LGTM!

Issue #515, opened 2022-01-14
Patch to starboard when 0.14.0 is out


NissesSenap:
Staboard v0.14.0-rc1 is currently out and when 0.14.0 get's released we should patch to the latest version.

We should set the following env variables: 
OPERATOR_VULNERABILITY_SCANNER_LATEST="true"
OPERATOR_VULNERABILITY_SCANNER_REPORT_TTL="24h"

They do exist as helm values.

This will create scans for only the latest RS and we will automatically run new scans after 24h.

NissesSenap:
Solved

PR #523, opened 2022-01-20
Update starboard to 0.14.0

NissesSenap:
* enable scanning the latest replicaset in a deployment.
* Set ttl of the vulnerability report to automatically be recreated.
* Solves: #515

### DIFF
--- a/modules/kubernetes/starboard/templates/starboard-values.yaml.tpl
+++ b/modules/kubernetes/starboard/templates/starboard-values.yaml.tpl
@@ -20,3 +20,7 @@ operator:
   configAuditScannerEnabled: false
   # kubernetesBenchmarkEnabled the flag to enable CIS Kubernetes Benchmark scanner
   kubernetesBenchmarkEnabled: false
+  # vulnerabilityScannerScanOnlyCurrentRevisions the flag to only create vulnerability scans on the current revision of a deployment.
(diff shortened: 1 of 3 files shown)


modules/kubernetes/starboard/templates/starboard-values.yaml.tpl
```diff
@@ -20,3 +20,7 @@ operator:
   configAuditScannerEnabled: false
   # kubernetesBenchmarkEnabled the flag to enable CIS Kubernetes Benchmark scanner
   kubernetesBenchmarkEnabled: false
+  # vulnerabilityScannerScanOnlyCurrentRevisions the flag to only create vulnerability scans on the current revision of a deployment.
+  vulnerabilityScannerScanOnlyCurrentRevisions: true
+  # vulnerabilityScannerReportTTL the flag to set how long a vulnerability report should exist. "" means that the vulnerabilityScannerReportTTL feature is disabled
+  vulnerabilityScannerReportTTL: "25h"
```
NissesSenap, line 26: How long TTL do we want?

NissesSenap:
As a part of enabling vulnerabilityScannerReportTTL you need to remove all the old vulnerability reports, else the old ones won't be automatically renewed.

```
kubectl delete vulnerabilityreports.aquasecurity.github.io -A --all
```

This will retrigger new scans.

PR #527, opened 2022-01-27
Add resource definition to grafana-agent-operator

jimgus:
The grafana-agent-operator Helm chart does not define any Kubernetes resource definitions by default. This will cause the Pod to be evicted. This PR adds a resource definition in order to avoid that problem.


modules/kubernetes/grafana-agent/charts/grafana-agent-extras/templates/grafanaagent.yaml
```diff
@@ -22,3 +22,9 @@ spec:
     instanceSelector:
       matchLabels:
         agent: grafana-agent-logs
+  resources:
```
NissesSenap, line 25: I would make these into helm variabels. Not that we should expose them as variabels in terraform. But it will make it easier to visualise all the values in one central file.

modules/kubernetes/grafana-agent/templates/operator-values.yaml.tpl
```diff
@@ -0,0 +1,9 @@
+resources:
+  requests:
+    cpu: 25m
+    memory: 100Mi
+tolerations:
+  - operator: "Exists"
```
NissesSenap, line 6: We don't need tolerations for the operator it self. We will always have standard nodes without tolerations that we can use.

modules/kubernetes/grafana-agent/charts/grafana-agent-extras/templates/grafanaagent.yaml
```diff
@@ -22,3 +22,9 @@ spec:
...
+    requests:
```
NissesSenap, line 26: I guess we skip limits due to that we can't configure the log agents and the metrics agents separately. Have you created any issue for this upstream? If so please link it and create a new issue in the terraform modules repo so we can keep a track of it.

modules/kubernetes/grafana-agent/templates/operator-values.yaml.tpl
```diff
@@ -0,0 +1,9 @@
+resources:
+  requests:
```
NissesSenap, line 2: Should be able to set limit on the operator. I haven't see the memory usage of the operator above 77Mi so I guess something like 120 should be more than enough.

NissesSenap (commented):
A few comments.
Don't forget to define resources for: modules/kubernetes/grafana-agent/charts/grafana-agent-extras/templates/traces-deployment.yaml

modules/kubernetes/grafana-agent/charts/grafana-agent-extras/templates/grafanaagent.yaml
jimgus: So far I have not seen any need for separate limits since VPA has recommended the same for both in the same environment. Let's do that when we see a need for it

NissesSenap: Since things can take a long time upstream I think it would be good to create the issue and I don't think this will be true in the long run. Especially in bigger clusters with lots and lots of metrics.

We can at least see what the upstream community thinks about it.

jimgus: https://github.com/grafana/agent/issues/1348

NissesSenap (dismissed):
We can fine tune limits in the long run. Let's get this merged for now.
LGTM

NissesSenap (approved):

PR #530, opened 2022-02-03
Add network ACLs to eks-core

jimgus:
The NLBs in front of the K8S cluster are running in the `public` and `private` subnets. In order restrict access to the NLBs, this PR adds Network ACLs to the subnets they are running in. The reason for using Network ACLs is that it is not possible to associate NLBs with security groups.


jimgus:
Let's do this after module refactoring.

PR #531, opened 2022-02-03
azure: Add prefix to role definitions

simongottschlag:
The azure_role_definition names needs to be unique for each Azure
tenant, which means if two different landing zones are using the same
everything else they will collide causing issues for the second one.

This fix should be backward compatible, only moving 'role' from being ha
hard coded string into a default variable that new landing zones will be
able to change if needed.

### DIFF
--- a/modules/azure/core/delegate-service-endpoint-join.tf
+++ b/modules/azure/core/delegate-service-endpoint-join.tf
@@ -1,5 +1,5 @@
 resource "azurerm_role_definition" "service_endpoint_join" {
-  name  = "role-${var.environment}-${var.location_short}-${var.name}-serviceEndpointJoin"
+  name  = "${var.azure_role_definition_prefix}-${var.environment}-${var.location_short}-${var.name}-serviceEndpointJoin"
   scope = azurerm_virtual_network.this.id
 
   permissions {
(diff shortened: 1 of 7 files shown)


modules/azure/core/delegate-service-endpoint-join.tf
```diff
@@ -1,5 +1,5 @@
 resource "azurerm_role_definition" "service_endpoint_join" {
-  name  = "role-${var.environment}-${var.location_short}-${var.name}-serviceEndpointJoin"
+  name  = "${var.azure_role_definition_prefix}-${var.environment}-${var.location_short}-${var.name}-serviceEndpointJoin"
```
phillebaba, line 2: Smart solution to keep backwards compatibility. Might be better in the long run to use unique suffix? If I remember correctly we already add an option to enable/disable this for keyvaults.

simongottschlag: @phillebaba We could use it, but since we are using azure_ad_group_prefix the same way as I'm implementing here, it seem like the logical way to continue with it. Suffix is in my opinion used for globally unique resources while prefix is used for tenant wide (groups, role definitions etc).

phillebaba: Sounds good to me.

phillebaba (approved):

PR #532, opened 2022-02-03
azure/core: Fix bug in route table association

simongottschlag:
This is **not** used by XKF by default and does not affect it if routes haven't been configured.

We were looping through var.routes which contains a list of all the
route+subnet combinations, which would work for the first route but not
for the second and third because the subnet already had a route table
associated.

With this fix, we are moving the configuration to use the correct
var.route_config instead which is the same as the route_tabe
configuration.

This shouldn't cause any big issue, but will temporarily remove the
route association from the subnet and then add it again. But after this
fix, we will be able to use multiple routes for a route table - as god
intended it to be. :^)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


simongottschlag:
@NissesSenap additional context added to description/commit message/changelog that this does **not** affect XKF by default, if custom rotues are not configured.

NissesSenap (approved):

NissesSenap (approved):

PR #533, opened 2022-02-04
Update cert manager to 1.6.1

jimgus:


modules/kubernetes/cert-manager/templates/values.yaml.tpl
```diff
@@ -20,3 +20,13 @@ webhook:
   securePort: 10260
   hostNetwork: true
 %{ endif }
+
```
NissesSenap, line 23: Set request for the main cert-manager oid as well.
Other then that it LGTM

NissesSenap (approved):
LGTM

PR #541, opened 2022-02-09
azure/core: add support to enable private endpoints on subnets

tvuorenmaa89:
This change will make it possible to enable private endpoints on specified subnets.
Current functionality will stay the same if the variable isn't configured.


NissesSenap:
Good job @tvuorenmaa89! 
Just run a `make docs` and push in those changes and the CI will be happy as well.

simongottschlag:
@phillebaba @NissesSenap Tobias and I did this one together, leaving review up to you. Tobias is currently installing the correct version of terraform-docs and will commit it.

NissesSenap (approved):
No ideal but it's a good workaround since we don't have defaults in objects.
LGTM what do you think @phillebaba ?

phillebaba:
There may be a reason for this, but why are we not adding the private endpoints to the subnet object instead?

phillebaba (dismissed):
This looks fine to me, however I will loop back to this implementation when optional object variables are implemented into Terraform.

simongottschlag:
Thank you @NissesSenap & @phillebaba!

@tvuorenmaa89 can you please rebase? I'll approve it when you're done. 😊

NissesSenap (dismissed):

NissesSenap (approved):

PR #545, opened 2022-02-14
Change prometheus volume size from 5Gi to 10Gi

phillebaba:
This change hopes to fix some issues with Prometheus volumes filling up during remote write down time.


NissesSenap (approved):
LGTM

PR #551, opened 2022-02-15
Fix pod label selector for Prometheus monitor

phillebaba:
Seems like the prometheus operator has removed the label `app: prometheus`. This changes fixes the label selector so that we get meta metrics.


NissesSenap (approved):
something something breaking changes....
LGTM

PR #552, opened 2022-02-15
Increase VolumeSize for Prometheus running in agent mode

CalleB3:
We need to set prometheus disk to 10Gi to lower the amount of issues we have with a growing WAL in our prometheus instances that is running in agent mode.


NissesSenap:
This is already solved https://github.com/XenitAB/terraform-modules/pull/545
Misscommunication between me and @CalleB3

PR #553, opened 2022-02-15
Remove secret and configmap data collection form kube state metrics

phillebaba:
This change removes the collection of secrets and configmaps from kube state metrics. My reasoning is that we do not use these metrics at all so there is no real point of collecting them. They only create a large amount of data which has to be remotely sent and received.


NissesSenap (approved):
Sounds reasonable.
LGTM

PR #558, opened 2022-02-17
Add ClusterRole for kubectl get nodes

CalleB3:
Adding ClusterRole to enable our users to use "kubectl get nodes"


modules/kubernetes/aks-core/k8s-cluster-role.tf
```diff
@@ -80,3 +80,17 @@ resource "kubernetes_cluster_role" "starboard_reports" {
     verbs      = ["get", "list", "watch", "update", "delete"]
   }
 }
+
+resource "kubernetes_cluster_role" "get-node" {
```
phillebaba, line 84: Need to snake case `_` resource names :)

CalleB3: Fixed :)

PR #560, opened 2022-02-18
Add SecretProviderClass CRD to ClusterRole

### LINKED ISSUE #559: Create clusterrole for CRD:s that we provide to tenants
We have a number of CRD in XKS. A number of these rules is something that our tenants should be able to use.

Go through the existing CRD:s that we have on all clusters and define them.
For example:

```.yaml
rules:
- apiGroups:
  - "secrets-store.csi.x-k8s.io"
  resources:
  - "*"
  verbs:
  - get
  - list
  - watch
```

Write a new rolebinding that is applied to each tenant.

Many of these role is already available through other clusterroles, for example certmanager. For now let's keep it that way.

CalleB3:
Add secretProviderClass CRD to existing clusterRole "custom_resource_edit"

Fixes #559


NissesSenap (approved):
Short and sweet

Issue #554, opened 2022-02-16
Set KubeletDiskType to Temporary to be able to use all disks on azure vm:s


NissesSenap:
Azure got a preview feature of being able to set the kubelet disk type to temporary. This configures the kubelet to use the temporary disk that is available on all azure vm:s. This disk is bigger in general compared to many OS disk or ephemeral disks that is included in azure vm:s.

For example we can see that a `Standard_D4a_v4`  https://docs.microsoft.com/en-us/azure/virtual-machines/dav4-dasv4-series#dav4-series got 100 GiB of SSD temp storage.
It would be great to store all logs and container images on this disk.

https://github.com/Azure/azure-rest-api-specs/blob/5582a35deb1bfa4aa22bac8f1d51b7934ead94ac/specification/containerservice/resource-manager/Microsoft.ContainerService/stable/2021-02-01/managedClusters.json#L3846-L3857

Sadly terraform currently don't support [kubelet_disk_type](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster#kubelet_disk_type)

We should do some research and see for how long this feature have been in preview and depending on that it might be worth to try to implement it in the azurerm provider.

NissesSenap:
Upstream issue created: https://github.com/hashicorp/terraform-provider-azurerm/issues/15449

NissesSenap:
https://github.com/Azure/AKS/releases/tag/2022-04-03

```
Custom node config  for AKS is now generally available.
https://docs.microsoft.com/azure/aks/custom-node-configuration
```

landerss1:
This attribute is available for the `default_node_pool` attribute in `azurerm_kubernetes_cluster` and in `azurerm_kubernetes_cluster_node_pool`.

landerss1:
Solved by [#1152](https://github.com/XenitAB/terraform-modules/pull/1153).

PR #561, opened 2022-02-21
Add kubelet_disk_type in AKS additional_node_pools

NissesSenap:
This allows us to use Temporary disk types which should remove most of
the risk of running ful on images, emptyDir and logs on AKS nodes.
Solves #554

Upgrade azurerm to 2.97 to get this new feature.

The valid options for `kubelet_disk_type` are: 
OS and Temporary


NissesSenap:
If you don't want to change your existing node pool all you need to do is to configure it to use: `kubelet_disk_type = "OS"`

NissesSenap:
We have read lots and lots of docs and we don't think that the kubelet_disk_type is in preview and thus ready for production.

https://github.com/Azure/azure-rest-api-specs/blob/main/specification/containerservice/resource-manager/Microsoft.ContainerService/stable/2022-01-01/managedClusters.json

For more information:
https://github.com/Azure/AKS/issues/930
https://docs.microsoft.com/en-us/rest/api/aks/agent-pools/create-or-update#kubeletdisktype

NissesSenap:
When trying to apply the config I get the error saying that it's a preview feature. So I think we can skip this for now since it will only force us to set another config value.

```
│ Error: creating Node Pool: (Agent Pool Name "standard3" / Managed Cluster Name "aks-dev-we-aks1" / Resource Group "rg-dev-we-aks"): containerservice.AgentPoolsClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="PreviewFeatureNotRegistered" Message="Preview feature Microsoft.ContainerService/KubeletDisk not registered."
│
```

NissesSenap:
As noted in the issue this feature have now gone GA.
An interesting thing to know is that even though terraform say it can do a in-place change it seems like it isn't possible.

I ran a in-place change but after running it a second time I get the following information. 

```
      ~ kubelet_disk_type      = "OS" -> "Temporary"
```

So my guess is that Azure changed back...
I'm currently trying to verify if this is the case when I create a new node pool or not.

NissesSenap:
I have misunderstood, I thought that custom node config ment kubelet disk type since it's a configuration that is done on the node... It's not.

Even though I can't find anything about kubelet disk type https://docs.microsoft.com/en-us/rest/api/aks/managed-clusters/create-or-update#kubeletdisktype being in preview still.
I might miss something obvious in the terraform provider but I'm unsure...

modules/azure/aks/variables.tf
```diff
@@ -36,15 +36,16 @@ variable "aks_config" {
...
-      min_count      = number
-      max_count      = number
-      spot_enabled   = bool
-      spot_max_price = number
-      node_taints    = list(string)
-      node_labels    = map(string)
+      name              = string
+      version           = string
+      vm_size           = string
+      kubelet_disk_type = string
```
NissesSenap, line 42: Should we have this as a temporary variable to be able to migrate easier between versions?
Or should we just remove it as a variable?

CHANGELOG.md
```diff
@@ -30,6 +30,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 - [#664](https://github.com/XenitAB/terraform-modules/pull/664) Manage Helm chart CRDs outside of Helm.
 - [#678](https://github.com/XenitAB/terraform-modules/pull/678) Update OPA to 3.8.1, gatekeeper-library to 0.12.1 and add k8srequireingressclass constraint.
+- [#561](https://github.com/XenitAB/terraform-modules/pull/561) [Breaking] Add kubelet_disk_type as a config to AKS additional_node_pools. this allows us to use `Temporary` disk types.
```
NissesSenap, line 33: Need to update changelog

NissesSenap (commented):
It's still a preview feature :( 

```shell
│ Error: creating Node Pool: (Agent Pool Name "standard1" / Managed Cluster Name "aks-qa-we-aks1" / Resource Group "rg-qa-we-aks"): containerservice.AgentPoolsClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="PreviewFeatureNotRegistered" Message="Preview feature Microsoft.ContainerService/KubeletDisk not registered."
```

landerss1:
Replaced by [#1152](https://github.com/XenitAB/terraform-modules/pull/1153).

PR #570, opened 2022-02-23
aks/eks core: fix network policy bug

### LINKED ISSUE #569: aks/eks core: fix default network policy
https://github.com/XenitAB/terraform-modules/pull/522 introduced network policies for commuinication to/from datadog and grafana-agent namespaces, but enabled even though default deny isn't enabled - causing these to be the only ones in the namespace and blocking all other traffic.

simongottschlag:
For cluster using var.kubernetes_network_policy_default_deny = false,
these new network policies are causing issues (all traffic is blocked).

Fixes #569

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## Unreleased
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


NissesSenap (approved):
This will fix the bug but is this the way we want to go?
It would be preferable to remove the variable and have everyone using networkpolicies. But assume this isn't a viable option right now.

So LGTM

simongottschlag:
@NissesSenap something for another issue and/or PR. :^) Thanks!

PR #572, opened 2022-02-23
Add storageClass in AKS to use ZRS disk.

NissesSenap:
This storageClass should only be used in specific use cases.
But it's very usefull when managing jobs over multiple nodes/zones.
ZRS=Zone Redudant Storage

For more information see: https://github.com/kubernetes-sigs/azuredisk-csi-driver/tree/master/deploy/example/topology#zrs-disk-support


phillebaba:
Rebase and we can merge

NissesSenap:
Have verified locally and it looks okay.

phillebaba (approved):

PR #574, opened 2022-02-23
Update cert-manager version

phillebaba:
The major change here is that all CRD versions other than v1 are removed. We need to do a quick inventory och clusters to determine that old versions are not being used. Probably not the case as v1 has been present for a while but we still need to make sure.


NissesSenap (approved):
What do we need to do to apply this change?
Just update and the CRD:s will be fixed?

I guess check which versions we are using currently in our clusters and make sure we update those accordingly.

phillebaba:
Yes I have checked most clusters and cant find anyone using anything else than v1. It wont stop current solutions from working, just stop future deployents if any non v1 api versions are used.

NissesSenap (approved):

PR #576, opened 2022-02-24
Upgrade gatekeeper version

phillebaba:


modules/kubernetes/opa-gatekeeper/templates/gatekeeper-values.yaml.tpl
```diff
@@ -1,7 +1,6 @@
 postInstall:
   labelNamespace:
     enabled: false
-experimentalEnableMutation: true
```
NissesSenap, line 4: is this the new default value?

NissesSenap (approved):

modules/kubernetes/opa-gatekeeper/templates/gatekeeper-values.yaml.tpl
phillebaba: Yes it has been removed for another value, mutations are now enabled by default.

NissesSenap (approved):
Isnt this a breaking change?

phillebaba:
No CRDs are managed by a job in the chart. Upgrade should be seamless.

Issue #550, opened 2022-02-14
grafana-agent loginstance config podLogsNamespaceSelector


NissesSenap:
We currently support datadog tenant users to be able to get nginx-ingress data.
Currently we enforce setting 

https://github.com/XenitAB/terraform-modules/blob/fbc773fb84de037c7efa692fda045739b66811df/modules/kubernetes/grafana-agent/charts/grafana-agent-extras/templates/loginstance.yaml#L21-L23

Since ingress-nginx got `xkf.xenit.io/kind: platform` we are unable to read that config.

* We need to come up with a more advanced label schema that can support both use cases.
* Then we need create a PR to support this in our LogsInstance & metrics config.
* Then create logscraping and metrics scraping by default/variable so we can enable it for our tenants.

NissesSenap:
This is solved.

PR #578, opened 2022-02-28
Add ingress-nginx metrics and logs scraping to grafana-agent

NissesSenap:
Solves #550


NissesSenap:
I have verified and I can see both metrics and logs from ingress-nginx controller in my grafana-agent instance.

phillebaba (approved):

PR #579, opened 2022-03-02
Add api & audit logging for eks

nissessenap:


NissesSenap:
This PR forces you to run

```shell
tf apply -var-file=variables/common.tfvars -var-file=variables/qa.tfvars -var-file=../global.tfvars -target='module.eks_global'
tf apply -var-file=variables/common.tfvars -var-file=variables/qa.tfvars -var-file=../global.tfvars -target='module.eks2'
```

This is due to the connection getting changed thanks to the update in eks cluster.

When these two runs have been made you will be able to run your standard `make plan` commands.

NissesSenap:
Error in the docs: https://docs.aws.amazon.com/AmazonCloudWatch/latest/logs/encrypt-log-data-kms.html also have to provide the region.

modules/aws/eks/eks.tf
```diff
@@ -51,6 +54,63 @@ resource "aws_eks_cluster" "this" {
...
+}
+
+
+
+# CloudWatch Encrytion
+resource "aws_kms_key" "cloudwatch_encryption" {
+  description         = "Used for cloudwatch encryption"
+  enable_key_rotation = true
+
+  policy = <<EOF
```
NissesSenap, line 68: This is a ugly workaround by AWS since you can't use a custom IAM role in your cloudWatch and you can't create the cloudWatch without having the IAM rule. 
But this only access for a specific cloudWatch group to the encryption from the specified ArnEquals.

This is the main reason why I use one KMS per EKS instance instead of storing it in global. If you think it's worth to move it back to global I will do that.

modules/aws/eks/eks.tf
```diff
@@ -33,10 +34,12 @@ data "aws_subnet" "cluster" {
 resource "aws_eks_cluster" "this" {
   provider = aws.eks_admin
 
-  name     = "${var.environment}-${var.name}${var.eks_name_suffix}"
+  name     = local.eks_cluster_name
   role_arn = var.cluster_role_arn
   version  = var.eks_config.kubernetes_version
 
+  enabled_cluster_log_types = ["api", "audit"]
```
NissesSenap, line 41: Do we want to make this into to a variable to make it easy to remove/add specific logs that you are interested in?

modules/aws/eks/variables.tf
```diff
@@ -62,3 +62,9 @@ variable "velero_config" {
     s3_bucket_arn = string
   })
 }
+
+variable "eks_cloudwatch_retention_period" {
+  description = "cloudwatch "
```
simongottschlag, line 67: extra space at the end

modules/aws/eks/eks.tf
```diff
@@ -51,6 +54,63 @@ resource "aws_eks_cluster" "this" {
...
   }
 
+  tags       = local.global_tags
+  depends_on = [aws_cloudwatch_log_group.this]
+}
+
+
+
+# CloudWatch Encrytion
+resource "aws_kms_key" "cloudwatch_encryption" {
```
simongottschlag, line 64: Should encryption be enabled by default?

modules/aws/eks/eks.tf
simongottschlag: What else is available?

modules/aws/eks/eks.tf
NissesSenap: I don't see why not. My understanding is that it doesn't come with any cost.

modules/aws/eks/eks.tf
NissesSenap: https://docs.aws.amazon.com/eks/latest/userguide/control-plane-logs.html

- Kubernetes API server component logs (api) 
- Audit
- authenticator
- Controller manager (controllerManager)
- Scheduler (scheduler)

NissesSenap:
Good documentation about AWS logging: https://aws.github.io/aws-eks-best-practices/security/docs/detective/

jimgus (commented):
Don't the CloudWatch part belong to the Global module? It would be nice to have the same log group for both clusters and when we switch clusters the log group for previous cluster will dissappear

modules/aws/eks/eks.tf
```diff
@@ -51,6 +54,63 @@ resource "aws_eks_cluster" "this" {
...
+
+  policy = <<EOF
+{
+  "Version" : "2012-10-17",
+  "Id" : "key-default-1",
+  "Statement" : [ {
+      "Sid" : "Enable IAM User Permissions",
+      "Effect" : "Allow",
+      "Principal" : {
+        "AWS" : "arn:aws:iam::${data.aws_caller_identity.current.account_id}:root"
```
bittrance, line 76: I think `data.aws_caller_identity.current.account_id` is the Terraform user's account, whereas according to  https://docs.aws.amazon.com/kms/latest/developerguide/key-policy-default.html it should be the account in which we are creating the key?

NissesSenap:
> Don't the CloudWatch part belong to the Global module? It would be nice to have the same log group for both clusters and when we switch clusters the log group for previous cluster will disappear

I think @jimgus  makes a great point. If I were an auditor I would want to see logs from the old cluster even if we did a blue green and created a new cluster.

So after this comment I agree we should have the log group inside the global module so we don't delete the old logs by mistake.

The problem is that eks don't provide any way of setting a custom name for your cloudWatch logs. Instead they have to be called `/aws/eks/<cluster-name>/cluster`
https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/eks_cluster#enabling-control-plane-logging
Thanks to this we would have to send in `eks_name_suffix` in to the global module.

```
module "eks_global" {
  source = "github.com/xenitab/terraform-modules//modules/aws/eks-global?ref=eks_audit"

  environment           = var.environment
  subscription_name     = var.azure_subscription_name
  name                  = var.name
  unique_suffix         = var.unique_suffix
  eks_group_name_prefix = var.aks_group_name_prefix
  namespaces            = var.namespaces
  eks_name_suffix       = [1,2] // This can of course be a variable
}
```

But this kind of breaks the thought around the global module. We could also create a new module called just to handle this. But it's still the same issue and I would prefer not have a bunch of small modules without a good reason.

What do you think @simongottschlag and @phillebaba?

modules/aws/eks/eks.tf
```diff
@@ -51,6 +54,63 @@ resource "aws_eks_cluster" "this" {
...
+    },
+    {
+      "Effect": "Allow",
+      "Principal": { "Service": "logs.${data.aws_region.current.name}.amazonaws.com" },
+      "Action": [
+        "kms:Encrypt*",
+        "kms:Decrypt*",
+        "kms:ReEncrypt*",
+        "kms:GenerateDataKey*",
+        "kms:Describe*"
```
bittrance, line 89: I imagine you copied these from AWS docs? AWS docs are infamous for crappy IAM practices. Among other things: there is only `kms:Encrypt` and `kms:Decrypt` so those stars are spurious. I suspect you can reduce `kms:Describe*` to `kms:DescribeKey`. (ReEncrypt* and GenerateDataKey* are legit.) Details here https://docs.aws.amazon.com/service-authorization/latest/reference/list_awskeymanagementservice.html#awskeymanagementservice-actions-as-permissions .

modules/aws/eks/eks.tf
```diff
@@ -51,6 +54,63 @@ resource "aws_eks_cluster" "this" {
...
+      ],
+      "Resource": "*",
+      "Condition": {
+        "ArnEquals": {
+            "kms:EncryptionContext:aws:logs:arn": "arn:aws:logs:${data.aws_region.current.name}:${data.aws_caller_identity.current.account_id}:log-group:/aws/eks/${local.eks_cluster_name}/cluster"
```
bittrance, line 94: Surely the EKS resource knows its ARN? Maybe I'm reading it wrong, but it doesn't look like a reference here would be cyclical?

bittrance (commented):
Random comments.

modules/aws/eks/eks.tf
NissesSenap: Yes they do. The problem is that I need the KMS to be created first.
Then the creation of CloudWatch
And finally the creation of EKS.

So I need  to know the EKS arn before it exist, but since I can't do that I use this ugly workaround that I found.

simongottschlag:
Great work @NissesSenap! :1st_place_medal: 

I've now gone through the PR, the comments and the documentation.

I'll try to gather my thoughts below:

- I would prefer to start with logging activiated for all cluster control plane log types, and remove if we don't see a point in having them. The only point in removing one would be if the cost of having it is high.
- I don't see the point in enabling encryption right now, better to keep the change small and to point and add encryption in the future if required. Less is more, KISS etc.
- I don't want us to start introducing the different clusters to global if not required. We've been able to keep it away so far and I would prefer to keep it like that.
- When reading the documentation, the `depends_on` for the EKS cluster isn't a requirement - just something that the terraform documentation does (and says that it may not be needed in your environment). I would prefer not to have it, if everything still works without it. BUT, I may be reading the docs the wrong way and the `aws_cloudwatch_log_group` is required to exist before enabling `enabled_cluster_log_types` on the cluster (which means we will need to have the explicit depends_on).

NissesSenap:
@simongottschlag  so how do you want to solve the depends_on?
Should we take the risk of by mistake deleting the logs when doing a blue green to a new cluster?

simongottschlag:
@NissesSenap I don't see any alternatives to it unfortunately (did a strike through of that bullet).

modules/aws/eks/eks.tf
bittrance: As it turns out, `data.aws_caller_identity.current.account_id` will be correct because there is no way to act in account 2 with credentials from account 1 except by assuming a role in account 2.

In this example, `sandbox` is a sub-account in my private organization. In the second step, aws-vault implicitly assumes a role and executes a command:
```sh
$aws-vault exec bittrance -- aws sts get-caller-identity
{
    "UserId": "AIDA<ID>",
    "Account": "<account-1>",
    "Arn": "arn:aws:iam::<account-1>:user/bittrance"
}

$ aws-vault exec sandbox -- aws sts get-caller-identity
{
    "UserId": "AROA<ID>:1646383552799375356",
    "Account": "<account-2>",
    "Arn": "arn:aws:sts::<account-2>:assumed-role/OrganizationAccountAccessRole/1646383552799375356"
}
```

NissesSenap:
After many ifs and buts we have decided the follow.

- Move the cloudWatch log group to the global module
- remove the kms config
- make EKS more opinionated by only supporting eks_name_suffix > 3

We will not create any strange implicit dependencies between global and eks modules.
As long as you don't do target applies (which would probably create issues for us no matter what) we assume that the cloudWatch in global will be created before EKS starts to create logs.

Creating a cloudWatch logGroup takes a few seconds while creating takes minutes.
If for some reason the cloudWatch logGroup is created after EKS then we will see an error in the logGroup creation so we won't miss that this have happened. 

If that is the case the user will have to import the cloudWatch logGroup to the current state and add a depends_on to make sure it won't happen in the future.

simongottschlag (approved):

PR #581, opened 2022-03-03
Add config to allow annotations and block default values

### LINKED ISSUE #580: ingress-nginx annotation-word-blocklist
Today if we disable `allow-snippet-annotations` we disable it all the way and we also have `annotation-value-word-blocklist` set.
And if allow-snippet-annotations is true we don't have any blocklist.
But what about when we want the blocklist?

We need to add a new variable that allows us to enable `allow-snippet-annotations` and have the default: `annotation-value-word-blocklist`. 

https://github.com/XenitAB/terraform-modules/blob/1409de7c3a4154fd758364b0ed008665c4371a84/modules/kubernetes/ingress-nginx/templates/values.yaml.tpl#L48-L51

The default values needs to be backwards compatible with the current settings.

CalleB3:
Add config for deny_default_annotation_blocker.

If allow_snippet_annotations & deny_default_annotation_blocker are true, use default block values.

Fixes #580


modules/kubernetes/ingress-nginx/templates/values.yaml.tpl
```diff
@@ -49,6 +49,10 @@ controller:
     %{~ if !allow_snippet_annotations ~}
     annotation-value-word-blocklist: load_module,lua_package,_by_lua,location,root,proxy_pass,serviceaccount,{,},',\
     %{~ endif ~}
+    deny-default-annotation-blocker: ${deny-default-annotation-blocker}
```
NissesSenap, line 52: Unsure why you add this config. Is it a part of the helm chart that is missing for some reason?

modules/kubernetes/ingress-nginx/templates/values.yaml.tpl
```diff
@@ -49,6 +49,10 @@ controller:
...
+    %{~ if allow_snippet_annotations && deny-default-annotation-blocker ~}
```
NissesSenap, line 53: Shouldn't we delete row 49 etc as well?  `%{~ if !allow_snippet_annotations ~}`
Else we will have the same value 2 times or am I reading wrong?

NissesSenap (commented):
Still some work needed I think. 
Also please fix the CI <3

CalleB3:
Pushed up some changes based on the feedback above, please check once more :)

validation/kubernetes/aks-core/main.tf
```diff
@@ -82,9 +82,10 @@ module "aks_core" {
   }
 
   ingress_config = {
-    http_snippet              = ""
-    public_private_enabled    = false
-    allow_snippet_annotations = false
+    http_snippet                    = ""
+    public_private_enabled          = false
+    allow_snippet_annotations       = false
+    deny_default_annotation_blocker = false
```
NissesSenap, line 88: Please update this setting in eks-core as well.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Added
 
+- [#581](https://github.com/XenitAB/terraform-modules/pull/581) Add config to allow annotations and block default values.
```
NissesSenap, line 11: This is not added.
This is changed or fixed. I would say more towards fixed.

phillebaba:
Just a quick question. Is there a case where we will be using all four different combinations of `allow_snippet_annotations` and `deny_default_annotation_blocker`. It seems like we are adding conditions which are not needed and just increases complexity.

CalleB3:
@phillebaba 
We only want `annotation-value-word-blocklist` to apply when both are true, in other words, we want to be able to use `allow_snippet_annotations` without the blocklist. And if `allow_snippet_annotations` is false, then we won't need `annotation-value-word-blocklist`.
We could make it so that `annotation-value-word-blocklist` applies when only `deny_default_annotation_blocker` is true, but then it could be applied without `allow_snippet_annotations`. Which would be unnecessary?
Short answer; Yes, we want this possibility.
Hope this explains some more

NissesSenap:
We have decided to be more opinionated. Instead of creating a new variable we will just fix the current if statment to actually use the block list when allow_snippet_annotations is true.
We will fix this in a new pr.

PR #582, opened 2022-03-04
Node local dns tenant network networkpolicy

NissesSenap:
After a few weeks of usage of node-local-dns we started to see issues with talking to the dns.
I don't know why this happens but after reading up on node-local-dns I understand why we need this rule.

This IP is very specific and I don't add any logic around it since it's the real IP of coreDNS service.


NissesSenap:
There is another IP used in AWS so that is why I use a variable in node-local-dns.

simongottschlag (approved):

NissesSenap:
Have verified in another cluster without node-local-dns and it works as expected.
I feel ready to merge :)

simongottschlag (approved):

PR #583, opened 2022-03-07
Use annotation-value-word-blocklist by default in Ingress-nginx

### LINKED ISSUE #580: ingress-nginx annotation-word-blocklist

CalleB3:
Fixes #580

### DIFF
--- a/modules/kubernetes/ingress-nginx/templates/values.yaml.tpl
+++ b/modules/kubernetes/ingress-nginx/templates/values.yaml.tpl
@@ -46,7 +46,7 @@ controller:
       ${http_snippet}
     %{~ endif ~}
     allow-snippet-annotations: ${allow_snippet_annotations}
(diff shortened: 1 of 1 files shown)


NissesSenap (approved):
LGTM

PR #585, opened 2022-03-07
Upgrade Azure AD provider to v2

### LINKED ISSUE #507: Update all modules to latest azuread provider
We need to move to the latest azuread provider which means we will have to also move from Azure AD Graph (deprecated) to Microsoft Graph.

Suggested tasks:

- [ ] Setup lab environment with all modules using Azure AD (with current setup)
- [ ] Create branch with the latest azure ad provider
- [ ] Fix all deprecations / provider changes
- [ ] Document service principal permission changes between Azure AD Grapgh and Microsoft Graph
- [ ] Document migration steps to start using a release with the new provider

simongottschlag:
This PR upgrades the Azure AD provider to v2 (1.6.0 -> 2.18.0) and fixes the different deprecations and changes related to it.

See [`AzureAD v2.0 and Microsoft Graph`](https://registry.terraform.io/providers/hashicorp/azuread/latest/docs/guides/microsoft-graph) for more details on the upgrade. The big difference is that we are moving away from the deprecated Azure AD Graph to the Microsoft Graph. This change requires us to update the service principals used in regards to their permissions:

>  - Add API permission `Application.ReadWrite.All` (`Microsoft Graph`) to the service principal(s)
>  - Remove the following from the state: module.governance_global.data.azuread_application.owner_spn (related [issue](https://github.com/hashicorp/terraform-provider-azuread/issues/541))
>	  ```shell
>	  make state-remove ENV=dev DIR=governance
>	  confirm: governance/dev
>	  regexp: data.azuread_application
>	  ```
>  - Run plan / apply, validate that only `random_password` resources are removed
>  - Remove service principal(s) from the `User administrator` role
>  - Remove the `Application.ReadWrite.All` (`Azure Active Directory Graph`) permission and admin consent from the service principal(s)	
>  - Remove the `Directory.ReadWrite.All` (`Azure Active Directory Graph`) permission and admin consent from the service principal(s) if it exists

Please note that the above is for `governance` and the same issue that required `state-remove` may occur with other modules using the data source `azuread_application`. The following error will be thrown if it's required:

> Error: .group_membership_claims: missing expected [

The other changes that have been made are:

- Stopped generating `random_password` for a couple of resources since the provider handles it for `azuread_application_password`
- Removed `value` from `azuread_application_password` since the provider handles it (and migrates it)
- Update resources using `random_password` that was used for `azuread_application_password` and instead use the `value` attribute from the `azuread_application_password`
- The `name` attribute has been changed for all resources and where used it has been replaced with `display_name`
- Set `security_enabled = true` on `azuread_group` resources

Fixes #507

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Fixed
 
 - [#587](https://github.com/XenitAB/terraform-modules/pull/587) Fix space error in grafana-agent podLogs
+- [#585](https://github.com/XenitAB/terraform-modules/pull/585) [Breaking] Upgrade Azure AD provider to v2
 
 ## 2022.03.1
 
--- a/modules/aws/eks-global/README.md
+++ b/modules/aws/eks-global/README.md
@@ -4,14 +4,14 @@
 |------|---------|
 | <a name="requirement_terraform"></a> [terraform](#requirement\_terraform) | 0.15.3 |
 | <a name="requirement_aws"></a> [aws](#requirement\_aws) | 3.63.0 |
-| <a name="requirement_azuread"></a> [azuread](#requirement\_azuread) | 1.6.0 |
+| <a name="requirement_azuread"></a> [azuread](#requirement\_azuread) | 2.18.0 |
 
 ## Providers
 
 | Name | Version |
 |------|---------|
 | <a name="provider_aws"></a> [aws](#provider\_aws) | 3.63.0 |
-| <a name="provider_azuread"></a> [azuread](#provider\_azuread) | 1.6.0 |
+| <a name="provider_azuread"></a> [azuread](#provider\_azuread) | 2.18.0 |
 
 ## Modules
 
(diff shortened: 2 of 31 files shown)


simongottschlag:
@XenitAB/xenit FYI

simongottschlag:
There may be more changes to other modules or permissions for other things we do, but this seems to be working for botht the `governance` modules. I've both tried the upgrade path (first with 1.6.0 then with this) as well as a complete destroy / create and it seems to be working as expected at least.

NissesSenap:
First of all, amazing job @simongottschlag.
I have pulled the branch and trying to look for things that you might have forgotten but I'm unable to find any.
This have been one of the tasks I haven't looked forward to this year but your good information make me feel that this should go without any issues.

In short LGTM

Looking forward to getting this PR merged and used in production :)

NissesSenap (dismissed):

phillebaba (approved):
Great job @simongottschlag :rose: 

I am fine with merging this now, as the manual intervention is not too complicated. Waiting longer is going to hurt as much as having to do the migration steps.

phillebaba (approved):

PR #588, opened 2022-03-09
Migrate AZDO & Github flux-modules to one module

### LINKED ISSUE #557: Put fluxcd-v2-azdo and fluxcd-v2-github in to a single module
Today XKF don't support fluxcd-v2 with both git data in both azure devops and github at the same time.
We have seen use cases for this and it would be good to create a new module that merges both of them together.

Deprecate the older folders after a few versions depending on how easy it will be to migrate without creating any real issues for the end users.

CalleB3:
Just a draft for @phillebaba

Fixes #557


validation/kubernetes/fluxcd-v2/main.tf
```diff
@@ -0,0 +1,16 @@
...
+  azure_devops_pat       = "123456"
+  azure_devops_org       = "foo"
+  azure_devops_proj      = "bar"
+  github_org             = "foo"
+  github_app_id          = 123
+  github_installation_id = 123
+  github_private_key     = "foo" #tfsec:ignore:general-secrets-no-plaintext-exposure
+  environment            = "dev"
+  cluster_id             = "foobar"
+  namespaces             = []
```
bittrance, line 15: I'm not sure exactly how relevant it is in this context, but generally speaking testing and example values should be unique so as to minimize the risk of letting letting in the pasta cat. For example, a copy-paste error might produce a mixup between `azure_devops_org` and `github_org` both of which is "foo" here.

PR #590, opened 2022-03-09
Drop CAP_SYS_ADMIN in default constraints

NissesSenap:
Drop CAP_SYS_ADMIN from all containers in the cluster just like NET_RAW.
Update grafana-library to version 0.10.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - [#617](https://github.com/XenitAB/terraform-modules/pull/617) Upgrade falco to 0.31.1
 - [#536](https://github.com/XenitAB/terraform-modules/pull/536) Update OPA Gatekeeper Helm charts
 - [#626](https://github.com/XenitAB/terraform-modules/pull/626) [Breaking] Add support for multiple DNS zones.
+- [#590](https://github.com/XenitAB/terraform-modules/pull/590) Drop CAP_SYS_ADMIN through OPA.
 
 ### Fix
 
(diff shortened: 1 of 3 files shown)


NissesSenap:
From seccomp gatekeeper-libary docs:

```.txt
            allowedProfiles:
              type: array
              description: >-
                An array of allowed profile values for seccomp on Pods/Containers.

                Can use the annotation naming scheme: `runtime/default`, `docker/default`, `unconfined` and/or
                `localhost/some-profile.json`. The item `localhost/*` will allow any localhost based profile.

                Can also use the securityContext naming scheme: `RuntimeDefault`, `Unconfined`
                and/or `Localhost`. For securityContext `Localhost`, use the parameter `allowedLocalhostProfiles`
                to list the allowed profile JSON files.

                The policy code will translate between the two schemes so it is not necessary to use both.

                Putting a `*` in this array allows all Profiles to be used.

                This field is required since with an empty list this policy will block all workloads.
              items:
                type: string
            allowedLocalhostFiles:
              type: array
              description: >-
                When using securityContext naming scheme for seccomp and including `Localhost` this array holds
                the allowed profile JSON files.

                Putting a `*` in this array will allows all JSON files to be used.

                This field is required to allow `Localhost` in securityContext as with an empty list it will block.
```

NissesSenap:
https://github.com/XenitAB/terraform-modules/pull/576 needs to be merged before this one got a chance.
Need to update CRD versions.

NissesSenap:
Don't know why I did it but seccomp should be done in a separate PR.

```
    {
      kind               = "K8sPSPSeccomp"
      name               = "psp-capabilities"
      enforcement_action = ""
      match = {
        kinds      = []
        namespaces = []
      }
      parameters = {
        allowedProfiles       = ["*"]
        allowedLocalhostFiles = ["*"]
      }
    },
```

NissesSenap:
Don't know why I did it but seccomp should be done in a separate PR.

```
    {
      kind               = "K8sPSPSeccomp"
      name               = "psp-capabilities"
      enforcement_action = ""
      match = {
        kinds      = []
        namespaces = []
      }
      parameters = {
        allowedProfiles       = ["*"]
        allowedLocalhostFiles = ["*"]
      }
    },
```

phillebaba (approved):
LGTM

PR #596, opened 2022-03-10
Update validation tools

phillebaba:


simongottschlag (approved):
Looks good! I think the next step is to create a container image that we can use both locally and in CI/CD to keep everything in sync. But that's another PR.

PR #598, opened 2022-03-10
Fix tfsec AWS security warnings

phillebaba:
This changes fixes the following TFSec warnings.
https://aquasecurity.github.io/tfsec/v1.8.0/checks/aws/autoscaling/enforce-http-token-imds/
https://aquasecurity.github.io/tfsec/v1.8.0/checks/aws/cloudwatch/log-group-customer-key/

The first one has to be verified that it actually works with IRSA.


landerss1:
Won't do. AWS support has been removed.

PR #601, opened 2022-03-16
Fix api-group for get nodes role

CalleB3:


NissesSenap (approved):
LGTM, but i thought that last time as well so who knows ;)

PR #610, opened 2022-03-22
Update node-local-dns dns config with prefetch cache config

NissesSenap:
Add prefetch config to allow automatic updates of well used endpoints.
For more information about the default values see: https://coredns.io/plugins/cache/

### DIFF
--- a/modules/kubernetes/node-local-dns/charts/node-local-dns/templates/nodelocaldns.yaml
+++ b/modules/kubernetes/node-local-dns/charts/node-local-dns/templates/nodelocaldns.yaml
@@ -52,6 +52,7 @@ data:
         cache {
                 success 9984 30
                 denial 9984 5
+                prefetch 3 60s 15%
         }
         reload
         loop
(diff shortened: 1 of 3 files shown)


bittrance (approved):
Seems reasonable.

modules/kubernetes/node-local-dns/charts/node-local-dns/templates/nodelocaldns.yaml
```diff
@@ -86,7 +86,10 @@ data:
         }
     .:53 {
         errors
-        cache 30
+        cache {
+                prefetch 10 1m 25%
+                serve_stale 10m
```
phillebaba, line 91: After some discussions we came to the conclusion that we should make this configuration evidence based. So lets start off with the default configuration for `serve_stale` and set only `prefetch 10`.

After running this for a while we can through metrics look at if we are for example serving very old stale DNS records or if it is a non issue. Same goes for prefetching.

phillebaba (approved):

phillebaba (approved):

PR #611, opened 2022-03-22
Add option to turn off ingress-nginx metrics in grafana

### LINKED ISSUE #609: Add feature to disable ingress-nginx metrics to grafana-agent
Today it's not possible to disable ingress-nginx logs or metrics.

https://github.com/XenitAB/terraform-modules/blob/main/modules/kubernetes/grafana-agent/charts/grafana-agent-extras/templates/monitors.yaml
https://github.com/XenitAB/terraform-modules/blob/main/modules/kubernetes/grafana-agent/charts/grafana-agent-extras/templates/pod-logs.yaml

Add a feature that enables this feature. Would an option be to use the `extra_namespaces`, parse it and look for ingress-nginx or something like that?

CalleB3:
Not fully complete

---Edit
Should be complete, me and @jimgus put our head togheter.
This will set `extra_namespaces` default to `ingress-nginx` which in turn will enable metrics/monitoring in grafana agent.
Setting `extra_namespaces` to an empty list will disable it. 

Fixes #609


modules/kubernetes/grafana-agent/charts/grafana-agent-extras/templates/monitors.yaml
```diff
@@ -1,3 +1,4 @@
+{{ if ingressNginx }}
```
CalleB3, line 1: Not sure in this,
With the for loop, if `extra_namespaces` contains `ingress-nginx` the bool will eq `true`
hence the config will apply? Will need to set it so it will use the pod-logs.yaml and monitors.yaml when it eq `false`.

modules/kubernetes/grafana-agent/charts/grafana-agent-extras/templates/pod-logs.yaml
```diff
@@ -1,3 +1,4 @@
+{{ if ingressNginx }}
```
CalleB3, line 1: Not sure in this,
With the for loop, if `extra_namespaces` contains `ingress-nginx` the bool will eq `true`
hence the config will apply? Will need to set it so it will use the pod-logs.yaml and monitors.yaml when it eq `false`.

modules/kubernetes/grafana-agent/templates/extras-values.yaml.tpl
```diff
@@ -6,3 +6,5 @@ remote:
   metricsUrl: ${remote_write_metrics_url}
   logsUrl: ${remote_write_logs_url}
   tracesUrl: ${remote_write_traces_url}
+
+ingressNginx: ${ingress_nginx_metrics}
```
NissesSenap, line 10: This isn't only for metrics, it's also for logs. I think I would call the variable ingress_nginx or if you don't like it maybe ingress_nginx_observability.

NissesSenap (commented):
A nice backwards compatible change, added a minor comments about variable naming.
@simongottschlag do you have any input on this?

modules/kubernetes/grafana-agent/templates/extras-values.yaml.tpl
CalleB3: Sure, we can change it, maybe `ingress_nginx_grafana` could be appropriate as well?

NissesSenap: We are already in grafana, I think I would go with ingress_nginx_observability since it's very generic and will help us if we want to use tracing in the future for example.

CalleB3: Changed it to `ingress_nginx_observability`

NissesSenap (approved):
LGTM

PR #618, opened 2022-03-27
Create new EKS and AKS node pools before deleting existing node pools

phillebaba:
This change will hopefully protect against situations where there are two node pools and someone attempts to replace the node pool in one shot. Normally Terraform will first delete the existing node pool which will start the eviction process before creating the new node pool. This is normally not an issue but not idea either. This change will make sure that there are nodes for pods to be evicted to before draining the old nodes.


simongottschlag (approved):
Smart 👌

PR #621, opened 2022-03-28
Lifecycle test

jimgus:


bittrance:
This looks like a PR that is incomplete or should not be merged. It would be good if the PR desc said what is going on.

PR #624, opened 2022-03-31
Add Promtail for platform logs

jimgus:
This PR will add Promtail for shipping platform logs to Loki for Azure. Promtail will be an opt-in feature to begin with, hence it will be disabled by default.  Promtail will be installed using this Helm chart: https://github.com/grafana/helm-charts/tree/main/charts/promtail Promtail will use the same client cert and key as Prometheus for Authentication. Promail is configurable to ignore logs from certain namespaces (typically tenant namespaces) and it also ignores all logs from the ingress-nginx namespace due to the large volume (might change to include a subset of the logs later on). This PR also contain some AWS code but that part has not been tested yet.


modules/kubernetes/promtail/templates/values.yaml.tpl
```diff
@@ -0,0 +1,164 @@
...
+priorityClassName: "platform-high"
+
+resources:
+  limits:
+    memory: 200Mi
+  requests:
+    cpu: 50m
+    memory: 100Mi
+ 
+  %{~ if provider == "azure" ~}
```
NissesSenap, line 47: extra " "

NissesSenap (approved):

PR #625, opened 2022-04-04
Cahnge cert-manager to http01 for acme

NissesSenap:
This to make cert-manager less complex and let ingress-nginx verify the acme endpoint
instead of using txt DNS solutions.


NissesSenap:
We will probably not use this solution since we can't control the DNS switch in a blue green scenario.

NissesSenap:
This is solved in: https://github.com/XenitAB/terraform-modules/pull/626 instead.

PR #626, opened 2022-04-04
Add support for multiple dns zones

NissesSenap:
To migrate to the new dns zone you need to do a state move.

## AKS

```shell
tf state mv 'module.aks_global.azurerm_dns_zone.this' 'module.aks_global.azurerm_dns_zone.this["dev.example.com"]'
```

We tried to solve this with the new fancy terraform move block but since it's a map and you can't use for loops within the move statement it's not possible to do.

## For EKS

```shell
tf state mv 'module.core.aws_route53_zone.this["zone"]' 'module.core.aws_route53_zone.this["dev.example.com"]'
```


NissesSenap:
I will keep on working on this for EKS tomorrow.

modules/kubernetes/aks-core/variables.tf
```diff
@@ -409,6 +409,11 @@ variable "promtail_config" {
   })
 }
 
+variable "ingress_healthz_dns_zone" {
+  description = "The ingress-healthz dns zone that should be used"
+  type        = string
+}
+
```
NissesSenap, line 416: Thinking about changing this to `main_dns_zone` since I need this variable in multiple locations in EKS. If so I think I should do the same in AKS.

modules/kubernetes/eks-core/variables.tf
```diff
@@ -142,7 +142,7 @@ variable "cert_manager_config" {
   description = "Cert Manager configuration"
   type = object({
     notification_email = string
-    dns_zone           = string
+    dns_zones          = list(string)
```
NissesSenap, line 145: Should we do the work of changing this variable to `dns_zones` vs `dns_zone` or not?
If so we should do the same in AKS.

NissesSenap (commented):
Added some comments to discuss around.

modules/kubernetes/aks-core/variables.tf
NissesSenap: We are using the first item in the dns_zone list as the "main" dns zone to use.

modules/kubernetes/eks-core/variables.tf
NissesSenap: Changed back to using dns_zone instead. Might be ugly but we are consistent. We can change this in a future change if we want.

NissesSenap:
@phillebaba we decide to change the AZAD usage of fqdn vs dns_zone[0] in a seperate change. This way we will be able to write better changelogs and it will be simpler to implement the changes in small batches.
+ readability :)

jimgus (approved):

PR #635, opened 2022-04-14
Upgrade azurerm provider to 3.1.0

phillebaba:
Major upgrade of the azurerm provider. Will include some parameter changes but should not affect anything outside the modules.

### DIFF
--- a/modules/azure/aks-global/outbound-ips.tf
+++ b/modules/azure/aks-global/outbound-ips.tf
@@ -1,8 +1,9 @@
 resource "azurerm_public_ip_prefix" "aks" {
-  count               = var.public_ip_prefix_configuration.count
+  count = var.public_ip_prefix_configuration.count
+
   name                = "pip-prefix-${var.environment}-${var.location_short}-${var.name}-aks-${count.index}"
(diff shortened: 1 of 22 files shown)


phillebaba:
Tested this with the governance, core, aks, and hub module. No major issues just Terraform reloading a bunch of new data to specific resources.

modules/azure/aks-global/outbound-ips.tf
```diff
@@ -1,8 +1,9 @@
 resource "azurerm_public_ip_prefix" "aks" {
-  count               = var.public_ip_prefix_configuration.count
+  count = var.public_ip_prefix_configuration.count
+
   name                = "pip-prefix-${var.environment}-${var.location_short}-${var.name}-aks-${count.index}"
   resource_group_name = data.azurerm_resource_group.this.name
   location            = data.azurerm_resource_group.this.location
   prefix_length       = var.public_ip_prefix_configuration.prefix_length
-  sku                 = "Standard"
```
phillebaba, line 7: This is the default value so no need to specify it.

jimgus (approved):

PR #636, opened 2022-04-14
Add node local dns to AWS and enable by default

phillebaba:
Node local dns should now be used everywhere as we have verified that it works. That is why its added to AWS and enabled by default.


NissesSenap (approved):
Have you verified the default IP on AWS together with the default deny networkspolicy? I can't remember how i solved that. Other then that LGTM

modules/kubernetes/eks-core/modules.tf
```diff
@@ -409,3 +409,15 @@ module "vpa" {
...
+
+module "node_local_dns" {
+  depends_on = [module.opa_gatekeeper, module.prometheus]
+
+  for_each = {
+    for s in ["node-local-dns"] :
+    s => s
+    if var.node_local_dns_enabled
+  }
+
```
NissesSenap, line 421: You need to overwrite the default ip, they differ between EKS and AKS in our current default setup.

```shell
$ kubectl get svc kube-dns -n kube-system -o jsonpath={.spec.clusterIP}
172.20.0.10
```

Please add:

dns_ip = "172.20.0.10"

NissesSenap (commented):
Need to change the internal dns_ip that EKS uses.
Minor description fix is needed.

modules/kubernetes/aks-core/namespace.tf
```diff
@@ -99,7 +99,7 @@ resource "kubernetes_network_policy" "tenant" {
       }
       to {
         ip_block {
-          cidr = "169.254.25.10/32"
```
NissesSenap, line 102: Strange that this have worked. Since it does my guess is that it's okay to delete. I will verify this.

NissesSenap: I think we can leave this for now and we can verify that it's okay to remove this block all together.

jimgus (approved):

PR #639, opened 2022-04-16
Create Azure AD Application for azad-kube-proxy using aks-global

simongottschlag:
This PR introduces a long awaited feature now enabled with using the latests terraform Azure AD provider (and with that, Microsoft Graph) where we are creating the Azure AD application used by azad-kube-proxy with terraform instead of manually.

You will now configure `aks-global` like this:
```terraform
module "aks_global" {
  source = "github.com/xenitab/terraform-modules//modules/azure/aks-global?ref=[ref]"

  [...]

  azad_kube_proxy_config = {
    cluster_name_prefix = "aks"
    proxy_url_override  = ""
  }
}
```

And then `aks-core` like this:
```terraform
module "aks1_core" {
  source = "github.com/xenitab/terraform-modules//modules/kubernetes/aks-core?ref=[ref]"

  [...]

  azad_kube_proxy_enabled = true
  azad_kube_proxy_config = {
    fqdn                  = "aks.${var.dns_zone}"
    azure_ad_group_prefix = var.aks_group_name_prefix
    allowed_ips           = var.aks_authorized_ips
    azure_ad_app          = module.aks_global.azad_kube_proxy.azure_ad_app
  }
}
```

This will also make it easy to change the name that shows up in the kubectl context, from `aks-[env]` to something like `customer-aks-[env]` if we want to by changing the `cluster_name_prefix` from `aks` to `customer-aks`. This may be a breaking change and should be tested carefully before doing.

Importing current Azure AD Application:
```shell
ENV=dev
AZ_APP_ID=$(az ad app list --display-name "aks-${ENV}" --query "[0].appId" -o tsv)
AZ_APP_PERMISSION_ID=$(az ad app show --id ${AZ_APP_ID} --output tsv --query "oauth2Permissions[0].id")
AZ_APP_OBJECT_ID=$(az ad app show --id ${AZ_APP_ID} --output tsv --query objectId)
AZ_SP_OBJECT_ID=$(az ad sp show --id $AZ_APP_ID --output tsv --query objectId)
APP_ROLE_ASSIGNMENT_ID=$(az rest --method GET --uri "https://graph.microsoft.com/v1.0/servicePrincipals/${AZ_SP_OBJECT_ID}/appRoleAssignments" --query "value[0].id" -o tsv)

terraform import -var-file=../global.tfvars -var-file=variables/common.tfvars -var-file=variables/${ENV}.tfvars module.aks_global.module.azad_kube_proxy.azuread_application.this $AZ_APP_OBJECT_ID
terraform import -var-file=../global.tfvars -var-file=variables/common.tfvars -var-file=variables/${ENV}.tfvars module.aks_global.module.azad_kube_proxy.azuread_service_principal.this $AZ_SP_OBJECT_ID
terraform import -var-file=../global.tfvars -var-file=variables/common.tfvars -var-file=variables/${ENV}.tfvars module.aks_global.module.azad_kube_proxy.azuread_application_pre_authorized.azure_cli ${AZ_APP_OBJECT_ID}/preAuthorizedApplication/04b07795-8ddb-461a-bbee-02f9e1bf7b46
terraform import -var-file=../global.tfvars -var-file=variables/common.tfvars -var-file=variables/${ENV}.tfvars module.aks_global.module.azad_kube_proxy.random_uuid.oauth2_permission_scope_user_impersonation $AZ_APP_PERMISSION_ID
terraform import -var-file=../global.tfvars -var-file=variables/common.tfvars -var-file=variables/${ENV}.tfvars module.aks_global.module.azad_kube_proxy.azuread_app_role_assignment.ms_graph_directory_read_all ${AZ_SP_OBJECT_ID}/appRoleAssignment/${APP_ROLE_ASSIGNMENT_ID}
```

### DIFF
--- a/modules/azure/aks-global/azad-kube-proxy.tf
+++ b/modules/azure/aks-global/azad-kube-proxy.tf
@@ -1,0 +1,57 @@
+locals {
+  azad_kube_proxy_url = var.azad_kube_proxy_config.proxy_url_override == "" ? "https://aks.${var.dns_zone[0]}" : var.azad_kube_proxy_config.proxy_url_override
+}
+
+resource "azuread_application" "azad_kube_proxy" {
+  for_each = {
+    for s in ["azad-kube-proxy"] :
+    s => s
+    if var.azad_kube_proxy_config.enabled
+  }
+
+  display_name = "${var.azad_kube_proxy_config.cluster_name_prefix}-${var.environment}"
+  identifier_uris = [
+    local.azad_kube_proxy_url
+  ]
+
+  api {
+    requested_access_token_version = 2
+  }
+
+  required_resource_access {
+    resource_app_id = "00000003-0000-0000-c000-000000000000" # Microsoft Graph
+
+    resource_access {
+      id   = "7ab1d382-f21e-4acd-a863-ba3e13f7da61" # Directory.Read.All
+      type = "Role"
+    }
+  }
+
+  tags = [
+    "azad-kube-proxy",
+    "cluster_name:${var.azad_kube_proxy_config.cluster_name_prefix}-${var.environment}",
+    "proxy_url:${local.azad_kube_proxy_url}",
+  ]
+}
+
+resource "azuread_application_password" "azad_kube_proxy" {
+  for_each = {
+    for s in ["azad-kube-proxy"] :
+    s => s
+    if var.azad_kube_proxy_config.enabled
+  }
+
+  application_object_id = azuread_application.azad_kube_proxy.object_id
+}
+
+resource "azuread_application_pre_authorized" "azad_kube_proxy_azure_cli" {
+  for_each = {
+    for s in ["azad-kube-proxy"] :
+    s => s
+    if var.azad_kube_proxy_config.enabled
+  }
+
+  application_object_id = azuread_application.azad_kube_proxy.object_id
+  authorized_app_id     = "04b07795-8ddb-461a-bbee-02f9e1bf7b46" # Azure CLI
+  permission_ids        = azuread_application.azad_kube_proxy.oauth2_permission_scope_ids
+}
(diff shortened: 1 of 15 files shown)


NissesSenap:
I read through it on the phone and my initial thought is that it looks great. But I need to read the docs to verify how some things work before giving a final review.

We should merge this no matter what but I'm thinking of how we should solve this for our AWS environments. 

On the top of me head i don't think we use aks-global in aws. But we do use governance.
Once again I'm on the phone but would an option be to fix this in the  governance module?
Then we would get the issue with not being able to get data between states in a smooth way.
Another option is of course that we create a new subfolder that we call something and use that in both AWS and Azure. Because aks-global contains a bunch of other things and we don't want to run on AWS. 


Any way awesome job @simongottschlag and thanks for picking it up, been a feature i have been looking forward to for some time.

simongottschlag:
@NissesSenap I don't think this should be included in governance, but rather place it in a separate Azure module that can be imported for both AKS and EKS when running them.

NissesSenap:
Yeah i also think that is the easiest way forward.
Now the hard thing of coming to with a name. It would be good if it's not called azad since we might use it for something else in the future. But probably something azure related, since it's in azure ;).

modules/azure/aks-global/azad-kube-proxy.tf
```diff
@@ -0,0 +1,69 @@
+data "azuread_application_published_app_ids" "well_known" {}
+resource "random_uuid" "azad_kube_proxy_oauth2_permission_scope_user_impersonation" {
+
+}
+resource "azuread_service_principal" "msgraph" {
+  application_id = data.azuread_application_published_app_ids.well_known.result.MicrosoftGraph
+  use_existing   = true
+}
+
+locals {
```
phillebaba, line 10: NIT: Put this on the top

modules/azure/aks-global/azad-kube-proxy.tf
```diff
@@ -0,0 +1,69 @@
+data "azuread_application_published_app_ids" "well_known" {}
```
phillebaba, line 1: NIT: New line between resource

phillebaba (dismissed):
Looks good! Really nice that we can get this working automatically.

simongottschlag:
@phillebaba @NissesSenap I've moved the main logic to `modules/azure-ad/azad-kube-proxy` and using it from `aks-global` and it will hopefully be easy to use with AWS as well when needed. What do you think?

PR #641, opened 2022-04-18
AKS Set kubelet config pod_max_pid

### LINKED ISSUE #640: AKS define default config max pid
Configuring kubelet config just became GA and it's now possible to define pod_max_pid (how many pids a single pod maximum can use).

The default max pid value on linux is: 32768 and I have verified on one our AKS node and Azure seems to use the default value.

By default in AKS pod_max_pid is set to 0 aka infinite. To hinder noisy neighbours we should set a decent default value so a single pod can't shutdown a whole node.

NissesSenap:
This will force recreation of the node pools.

If we just focus on the node pool and not the default node pool I would recommend to add a new node pool in your var file.
Then I would do a targeted apply on that node pool.
When it's created you should be able to run a new plan and destroy the old node pool.

```shell
terraform apply -var-file=variables/common.tfvars -var-file=variables/sand.tfvars -var-file=../global.tfvars -target='module.aks1.azurerm_kubernetes_cluster_node_pool.this["standard3"]'
# Make sure that the new node pool is up and running.
# Run a new plan on the entire module and let TF destroy the old node pool
```

Fixes #640

Not visible in the portal but if you use the json view you can see

```
        "kubeletConfig": {
            "cpuCfsQuota": false,
            "failSwapOn": false,
            "podMaxPids": 5000
        },
```


NissesSenap:
We need to update azurerm to not force this as a preview feature.

```shell
│ Error: creating Node Pool: (Agent Pool Name "standard3" / Managed Cluster Name "aks-sand-we-aks1" / Resource Group "rg-sand-we-aks"): containerservice.AgentPoolsClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="PreviewFeatureNotRegistered" Message="Preview feature Microsoft.ContainerService/CustomNodeConfigPreview not registered."
```

NissesSenap:
Hurray this feature is finally available and I don't have to use it in preview mode :)

modules/azure/aks/aks.tf
```diff
@@ -59,6 +59,10 @@ resource "azurerm_kubernetes_cluster" "this" {
     enable_auto_scaling          = false
     node_count                   = var.aks_config.production_grade ? 2 : 1
     only_critical_addons_enabled = true
+
+    #kubelet_config {
+    #  pod_max_pid = 0
+    #}
```
phillebaba, line 65: Remove or uncomment.

NissesSenap: Do we think it's worth doing?
This will force a new creation of the default node pool.

I wanted to verify without having to do the migration and that is why i commented out.

jimgus (approved):

PR #645, opened 2022-04-21
Refactor AKS node configuration

phillebaba:
This change includes a couple of breaking changes.
* Validation of node names.
* Forcing the use of ephemeral nodes.
* Removes option to set OS disk size.
* Sku tier and default node pool size is decided on production grade setting.


modules/azure/aks/aks.tf
```diff
@@ -53,30 +53,20 @@ resource "azurerm_kubernetes_cluster" "this" {
   default_node_pool {
     name           = "default"
     vnet_subnet_id = data.azurerm_subnet.this.id
+    zones          = ["1", "2", "3"]
 
-    type                         = "VirtualMachineScaleSets"
```
phillebaba, line 57: This is the default

modules/azure/aks/aks.tf
```diff
@@ -85,15 +75,13 @@ resource "azurerm_kubernetes_cluster_node_pool" "this" {
...
-
-  os_disk_type         = each.value.os_disk_type
-  os_disk_size_gb      = each.value.os_disk_size_gb
-  orchestrator_version = each.value.orchestrator_version
+  enable_auto_scaling  = true
+  os_disk_type         = "Ephemeral"
+  os_disk_size_gb      = local.vm_skus_disk_size_gb[each.value.vm_size]
+  orchestrator_version = each.value.version
   vm_size              = each.value.vm_size
-  node_count           = each.value.min_count
```
phillebaba, line 96: Seems to be no reason to set this.

modules/azure/aks/aks.tf
```diff
@@ -102,4 +90,65 @@ resource "azurerm_kubernetes_cluster_node_pool" "this" {
 
   node_taints = each.value.spot_enabled ? concat(each.value.node_taints, ["kubernetes.azure.com/scalesetpriority=spot:NoSchedule"]) : each.value.node_taints
   node_labels = each.value.spot_enabled ? merge(each.value.node_labels, { "node-pool" = each.value.name, "kubernetes.azure.com/scalesetpriority" = "spot" }) : merge(each.value.node_labels, { "node-pool" = each.value.name })
+
+  upgrade_settings {
+    max_surge = "33%"
```
phillebaba, line 95: Based on recommendation from Azure docs.
https://docs.microsoft.com/en-us/azure/aks/upgrade-cluster

modules/azure/aks/variables.tf
```diff
@@ -26,27 +26,52 @@ variable "aks_name_suffix" {
 variable "aks_config" {
   description = "The Azure Kubernetes Service (AKS) configuration"
   type = object({
-    kubernetes_version = string
-    sku_tier           = string
+    version          = string
+    production_grade = bool
```
phillebaba, line 30: Need to find a better name for this.

modules/azure/aks/aks.tf
```diff
@@ -53,30 +53,22 @@ resource "azurerm_kubernetes_cluster" "this" {
...
-
-    orchestrator_version = var.aks_config.default_node_pool.orchestrator_version
-    node_count           = 1
+    orchestrator_version = var.aks_config.version
     # This is a bug in tflint
     # tflint-ignore: azurerm_kubernetes_cluster_default_node_pool_invalid_vm_size
-    vm_size = "Standard_D2as_v4"
-
-    node_labels = var.aks_config.default_node_pool.node_labels
+    vm_size                      = "Standard_D2ds_v5"
```
phillebaba, line 61: Azure docs wants larger nodes but I do not think we need them if we are using node local dns as the load on coredns will be a lot lower.
https://docs.microsoft.com/en-us/azure/aks/use-system-pools

NissesSenap, line 61: Wrote the same thing in the other PR. 
What happens if this vm type runs out? I don't want to be forced to cut a new release to solve that at 3 in the night to get a cluster up and running.

I would like to see this as a variable with Standard_D2ds_v5 as the default value.

NissesSenap (commented):
One minor comment. I realize now that I reviewed the PR:s in the wrong order.

modules/azure/aks/aks.tf
phillebaba: Dont see how this an issue as as we are not provisioning default node pools late at night. We have so far never experienced an issue with the general VM instances and cannot see how we will in the future. I would rather not make this configurable but if it is an issue for you I am fine with changing it.

phillebaba:
So if we can resolve the default node pool discussion I want to chain this PR with two more. One is the PR that adds Kubernetes version validation. The other is a PR which aligns EKS and AKS node configuration with each other.

modules/azure/aks/aks.tf
NissesSenap: That is very true at the same time we could definitely do a blue green at the middle of the night.
Yeah I think it would feel better for me if I knew I could overwrite this.
@jimgus we are talking about this in this PR instead.

modules/azure/aks/variables.tf
phillebaba: We will go with this name.

NissesSenap (approved):

PR #648, opened 2022-04-21
Exclude namespaces from Datadog

jimgus:
Prior to this change, when enabled, Datadog is configured to ship logs
from all tenant namespace plus a configured number of "extra" namespaces
through the "extra_namespaces" variable in eks_core and aks_core.
Now we also must be able to exclude namespaces from the tenant
namespaces. Instead of adding a "exclude_namespaces" variable, the
module interface is changed by removing the "extra_namespaces" variable
and an new "namespaces" variable is introduced where you configure all
namespaces that Datadog shall ship logs from explicitly. In this way
the module interface will be easier to understand compared to the other
alternative mentioned above. Also what namespace that are being used 
will be more explicit though the configuration of the module.

Previous usage example:

` extra_namespaces = ["ingress-nginx"]`

Will have to be changed to:

` namespaces = concat(var.namespaces[*].name, ["ingress-nginx"])`


NissesSenap:
No comments on the code itself. Could you provide a example what variables to send in to the module when using it?

NissesSenap (approved):

PR #651, opened 2022-04-22
Add new default assigns to OPA lib

NissesSenap:
* pod-default-seccomp
* disable default mount of SA token


phillebaba:
Rebase and add changelog and we should be fine with this.

PR #653, opened 2022-04-23
Add check for Kubernetes version

phillebaba:
Depends on  #645

### DIFF
--- a/modules/aws/eks/variables.tf
+++ b/modules/aws/eks/variables.tf
@@ -42,6 +42,20 @@ variable "eks_config" {
       max_size       = number
     }))
   })
+
+  validation {
+    condition = alltrue([
+      for np in concat(var.eks_config.node_pools, [{ version : var.eks_config.version }]) : can(regex("^1.(20|21|22)", np.version))
+    ])
+    error_message = "The Kubernetes version has not been validated yet, supported versions are 1.20, 1.21, 1.22."
+  }
+
+  validation {
+    condition = alltrue([
+      for np in var.eks_config.node_pools : split(".", np.version)[1] <= split(".", var.eks_config.version)[1] && split(".", split("-", np.version)[0])[2] <= split(".", var.eks_config.version)[2]
+    ])
+    error_message = "The node Kubernetes version should not be newer than the cluster version, upgrade the cluster first."
+  }
 }
 
 variable "cluster_role_arn" {
--- a/modules/azure/aks/variables.tf
+++ b/modules/azure/aks/variables.tf
@@ -50,6 +50,20 @@ variable "aks_config" {
 
   validation {
     condition = alltrue([
+      for np in concat(var.aks_config.node_pools, [{ version : var.aks_config.version }]) : can(regex("^1.(20|21|22)", np.version))
+    ])
+    error_message = "The Kubernetes version has not been validated yet, supported versions are 1.20, 1.21, 1.22."
+  }
+
+  validation {
+    condition = alltrue([
+      for np in var.aks_config.node_pools : split(".", np.version)[1] <= split(".", var.aks_config.version)[1] && split(".", np.version)[2] <= split(".", var.aks_config.version)[2]
+    ])
+    error_message = "The node Kubernetes version should not be newer than the cluster version, upgrade the cluster first."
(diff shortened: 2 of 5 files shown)


modules/aws/eks/variables.tf
```diff
@@ -32,16 +32,30 @@ variable "eks_authorized_ips" {
...
+      max_size       = number
+      instance_types = list(string)
     }))
   })
+
+  validation {
+    condition = alltrue([
+      for np in concat(var.eks_config.node_groups, [{ version : var.eks_config.version }]) : can(regex("^1.(19|20|21)", np.version))
+    ])
+    error_message = "The Kubernetes version has not been validated yet, supported versions are 1.19, 1.20, 1.21."
```
phillebaba, line 50: Sadly it is not possible to use locals in validation so this needs to be hardcoded.

modules/azure/aks/aks.tf
```diff
@@ -53,30 +53,22 @@ resource "azurerm_kubernetes_cluster" "this" {
...
-
-    orchestrator_version = var.aks_config.default_node_pool.orchestrator_version
-    node_count           = 1
+    orchestrator_version = var.aks_config.version
     # This is a bug in tflint
     # tflint-ignore: azurerm_kubernetes_cluster_default_node_pool_invalid_vm_size
-    vm_size = "Standard_D2as_v4"
-
-    node_labels = var.aks_config.default_node_pool.node_labels
+    vm_size                      = "Standard_D2ds_v5"
```
NissesSenap, line 61: Do we want to hardcode this value?
What happens if that node type runs out on AKS?
I still think it's a good idea to allow changing this value but use the suggested one as a default value in a variable.

NissesSenap (commented):
Really nice job, looking forward to using this.
Ether in https://github.com/XenitAB/terraform-modules/pull/645 or in this PR we need to write instructions on how to migrate to this new version.

I also think that before merging this we should cut a new release so we don't have to solve this to fix: https://github.com/XenitAB/terraform-modules/pull/656

Added a few comments

modules/azure/aks/aks.tf
```diff
@@ -53,30 +53,22 @@ resource "azurerm_kubernetes_cluster" "this" {
   default_node_pool {
     name           = "default"
     vnet_subnet_id = data.azurerm_subnet.this.id
+    zones          = ["1", "2", "3"]
 
-    type                         = "VirtualMachineScaleSets"
```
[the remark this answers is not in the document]
NissesSenap: This is mentioned in #645 .

modules/azure/aks/variables.tf
```diff
@@ -26,27 +26,66 @@ variable "aks_name_suffix" {
 variable "aks_config" {
   description = "The Azure Kubernetes Service (AKS) configuration"
   type = object({
-    kubernetes_version = string
-    sku_tier           = string
+    version          = string
+    production_grade = bool
```
jimgus, line 30: Would be nice to have a description on the semantics of "production_grade"

modules/azure/aks/aks.tf
jimgus: Fair point

jimgus (commented):
Great work!

modules/azure/aks/variables.tf
```diff
@@ -26,27 +26,66 @@ variable "aks_name_suffix" {
...
+  }
+
+  validation {
+    condition = alltrue([
+      for np in var.aks_config.additional_node_pools : can(regex("^[a-z0-9]+$", np.name))
+    ])
+    error_message = "The name value has to be lowercase alphanumeric."
+  }
+
+  validation {
```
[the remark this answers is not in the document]
phillebaba: No two separate things. One checks that all characters are a-z or 0-9. The other makes sure that the name starts with a-z. Weird Azure thing but you can have numbers in the name but the name cannot start with a number.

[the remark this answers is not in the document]
NissesSenap: Ahh got it, I thought you wanted to make sure that it was lover letters two times.

NissesSenap (approved):

PR #664, opened 2022-05-04
Manage Helm chart CRDs outside of Helm

phillebaba:
I have on purpose skipped Linkerd as I will solve it in a separate PR when I bump the version.

### DIFF
--- a/modules/kubernetes/aad-pod-identity/main.tf
+++ b/modules/kubernetes/aad-pod-identity/main.tf
@@ -16,6 +16,10 @@ terraform {
       source  = "hashicorp/helm"
       version = "2.4.1"
     }
+    kubectl = {
+      source  = "gavinbunney/kubectl"
+      version = "1.14.0"
+    }
   }
 }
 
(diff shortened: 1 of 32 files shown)


modules/kubernetes/aad-pod-identity/main.tf
```diff
@@ -29,8 +33,33 @@ resource "kubernetes_namespace" "this" {
...
+  }
+  server_side_apply = true
+  apply_only        = true
+  yaml_body         = each.value
+}
+
 #tf-latest-version:ignore
 resource "helm_release" "aad_pod_identity" {
+  depends_on = [kubectl_manifest.aad_pod_identity]
+
```
NissesSenap, line 62: Don't we need the skip_crds on aad-pod-identity as well?

modules/kubernetes/datadog/main.tf
```diff
@@ -44,7 +48,32 @@ resource "kubernetes_namespace" "this" {
   }
 }
 
+data "helm_template" "datadog_operator" {
+  repository   = "https://helm.datadoghq.com"
+  chart        = "datadog-operator"
+  name         = "datadog-operator"
+  version      = "0.7.0"
+  api_versions = ["apiextensions.k8s.io/v1/CustomResourceDefinition"]
```
NissesSenap, line 56: Such a nice workaround :)

modules/kubernetes/opa-gatekeeper/main.tf
```diff
@@ -118,13 +122,39 @@ resource "kubernetes_namespace" "this" {
...
+  name         = "gatekeeper"
+  version      = "3.7.1"
+  include_crds = true
+}
+
+data "kubectl_file_documents" "gatekeeper" {
+  content = data.helm_template.gatekeeper.manifest
+}
+
+resource "kubectl_manifest" "gatekeeper" {
```
NissesSenap, line 137: Don't gatekeeper manage the crd on it's own?

modules/kubernetes/cert-manager/main.tf
```diff
@@ -34,13 +38,42 @@ resource "kubernetes_namespace" "this" {
...
+    name  = "installCRDs"
+    value = "true"
+  }
+}
+
+data "kubectl_file_documents" "cert_manager" {
+  content = data.helm_template.cert_manager.manifest
+}
+
+resource "kubectl_manifest" "cert_manager" {
```
NissesSenap, line 56: Don't cert-manager manage the crd on it's own?

NissesSenap (approved):
Minor comments, if you don't think it's a problem go for it :)

modules/kubernetes/opa-gatekeeper/main.tf
phillebaba: I want to manage the CRDs in the same way for all charts, even if they have their own method to update CRDs

NissesSenap:
Output from a clean cluster:

```.hcl
│ Error: Invalid for_each argument
│ 
│   on .terraform/modules/aks1_core/modules/kubernetes/aad-pod-identity/main.tf line 49, in resource "kubectl_manifest" "aad_pod_identity":
│   49:   for_each = {
│   50:     for k, v in data.kubectl_file_documents.aad_pod_identity.manifests :
│   51:     k => v
│   52:     if can(regex("^/apis/apiextensions.k8s.io/v1/customresourcedefinitions/", k))
│   53:   }
│     ├────────────────
│     │ data.kubectl_file_documents.aad_pod_identity.manifests is a map of string, known only after apply
│ 
│ The "for_each" value depends on resource attributes that cannot be determined until apply, so Terraform cannot predict how many instances will be created. To work
│ around this, use the -target argument to first apply only the resources that the for_each depends on.
...
etc
``` 

I will give it a try with depends on.

NissesSenap (approved):
LGTM, great job!

PR #666, opened 2022-05-05
Enable ingress-nginx logs in promtail

jimgus:
Before this change, all logs from the ingress-nginx namespace were
dropped due to the massive amount of logs. This change will enable
ingress-nginx logging but drop the 2xx and 3xx entries from the
access log.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## Unreleased
 
+### Changed
+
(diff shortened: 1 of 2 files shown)


jimgus:
The log entries to be dropped from the Nginx access log looks like this in our setup:

`51.105.169.12 - - [04/May/2022:13:12:50 +0000] "POST /api/v1/receive HTTP/2.0" 200 0 "-" "Prometheus/2.35.0" 35234 0.004 [monitor-router-receiver-remote-write] [] 10.244.34.149:19291 0 0.004 200 0156a072a34b23dc09bcfcfe87991c7b`

NissesSenap (dismissed):

NissesSenap (approved):

PR #667, opened 2022-05-06
Ingress-nginx opentelemetry

NissesSenap:
Let's see if we can enable tracing in ingress-nginx.
Relates to #668


NissesSenap:
I close this for now. I'm following the upstream issues that exist and when it's solved I will pick this up again.

PR #669, opened 2022-05-11
Add customTemplate for ingrss-nginx

NissesSenap:
More info can be found:
https://github.com/nginxinc/kubernetes-ingress/tree/v2.2.0/examples/custom-templates
https://github.com/kubernetes/ingress-nginx/blob/main/rootfs/etc/nginx/template/nginx.tmpl


NissesSenap:
When trying to use the template I get:

```
F0511 08:55:00.690888       8 nginx.go:167] Invalid NGINX configuration template: template: nginx.tmpl:2: undefined variable "$server"
```

NissesSenap:
If we decide to go with default certificate in ingress-nginx this PR will take over instead: https://github.com/XenitAB/terraform-modules/pull/670

NissesSenap:
We can save the branch for now if we want to go towards using templates instead of default certificate.

PR #670, opened 2022-05-11
Ingress-nginx enable default certificate

NissesSenap:
This to pass PCI scanning without any issues.


simongottschlag (approved):
LGTM 👍

phillebaba (approved):
I will approve this but I really think that in the future we should look into disabling the default cert and having endpoints fail instead.

PR #671, opened 2022-05-12
Breaking governance move regional delegate-se to global module

NissesSenap:
During plan/apply the membership of the service endpoint will be destroyed and then recreated.
This should have minimal impact.

This to be able to have 2 regional resources at the same time in the governance module


CalleB3 (approved):
LGTM

PR #674, opened 2022-05-13
[BREAKING] Multi region support

CalleB3:
Closing this PR, 

We decided to start from scratch, some too quick decisons were made in this PR.


modules/azure/aks-global/aad-group.tf
```diff
@@ -1,35 +1,35 @@
 resource "azuread_group" "view" {
   for_each     = { for ns in var.namespaces : ns.name => ns }
-  display_name = "${var.aks_group_name_prefix}${var.group_name_separator}${var.subscription_name}${var.group_name_separator}${var.environment}${var.group_name_separator}${each.value.name}${var.group_name_separator}view"
+  display_name = "${var.aks_group_name_prefix}-${var.subscription_name}-${var.environment}-${each.value.name}-view"
```
simongottschlag, line 3: @CalleB3 why are we removing the variable for the separator?

simongottschlag (changes requested):

modules/azure/aks-global/aad-group.tf
jimgus: I think it is a good idea to remove it unless we really need to be able to change the separator. Improves readability a lot.

simongottschlag: @jimgus quite sure it's being used.

jimgus: @simongottschlag  Oh, that is a pity in that case

PR #676, opened 2022-05-17
Add AKS audit logs for eventhub

jimgus:
Enable AKS resource logs to our clusters.

Currently we are not interested in host monitoring, we just want the AKS resource logs https://docs.microsoft.com/en-us/azure/aks/monitor-aks#collect-resource-logs.

This PR uses a log RG where we for now will focus on storing AKS controller logs but we could store other logs here as well.
We use a seperate RG to minimize access to the logs that we want to store long term. 

We use a storage account for long term storage of audit logs while we use log-analytics for short term logs.

Add the following to your main.tf file under the aks module:
```
  log_analytics_id                     = module.aks_global.log_analytics_id
  log_storage_account_id        = module.aks_global.log_storage_account_id
```

This might also be needed.
```
  log_eventhub_name                  = module.aks_global.log_eventhub_name
  log_eventhub_authorization_rule_id = module.aks_global.log_eventhub_authorization_rule_id
```

Add the following to your governance common.tfvars

```.tfvars
  {
    common_name                = "log",
    delegate_aks               = false,
    delegate_key_vault         = false,
    delegate_service_endpoint  = false,
    delegate_service_principal = false,
    lock_resource_group        = true,
    tags = {
      "description" = "Managing logs"
    }
  },
```


modules/azure/aks/aks.tf
```diff
@@ -62,6 +62,56 @@ resource "azurerm_kubernetes_cluster" "this" {
...
+    }
+  }
+
+  log {
+    category = "kube-audit"
+    enabled  = true
+
+    retention_policy {
+      enabled = true
+      days    = 30
```
NissesSenap, line 105: The retention time for the audit log is rather low.
One option forward could be to have two azurerm_monitor_diagnostic_setting definitions.
One where we send data to log analytics and the second where we send the data to a storage account.

This would lower the cost of the aduit logs by allot. Then we would of course have to create a good way of working of easily reading out this data.

Another option could be to have a event-hub in between get all the data, send some to log analytics and some to storage account.

And a third option is to send all the data to a storage account but then we have to be really good at reading the data from the storage account.

modules/azure/aks-global/log-analytics-workspace.tf
```diff
@@ -0,0 +1,20 @@
+resource "azurerm_log_analytics_workspace" "aks" {
+  name                = "aks-${var.environment}-${var.location_short}-${var.name}-${var.unique_suffix}"
+  location            = data.azurerm_resource_group.this.location
+  resource_group_name = data.azurerm_resource_group.this.name
+  sku                 = "PerGB2018"
+  retention_in_days   = 30
+}
+
+resource "azurerm_log_analytics_solution" "logs" {
```
NissesSenap, line 9: This is a pre-built solution that Azure provides: https://docs.microsoft.com/en-us/azure/azure-monitor/containers/container-insights-overview

NissesSenap: Removed this since we don't gather any vm info.

modules/azure/aks-global/logs.tf
```diff
@@ -0,0 +1,18 @@
+resource "azurerm_log_analytics_workspace" "log" {
+  name                = "log-${var.environment}-${var.location_short}-${var.name}-${var.unique_suffix}"
+  resource_group_name = data.azurerm_resource_group.log.name
+  location            = data.azurerm_resource_group.log.location
+  sku                 = "PerGB2018"
+  retention_in_days   = 30
+}
+
+#tfsec:ignore:azure-storage-queue-services-logging-enabled
+resource "azurerm_storage_account" "log" {
```
NissesSenap, line 10: Do we want to manage these logs in aks-global.
It might be better to move them to core.

My thinking is that if we use core it will become a more natural part of sending all of our logs to.
No matter if it's key vault audit logs or AKS audit logs.

What do you think?

NissesSenap:
This was partly solved in https://github.com/XenitAB/terraform-modules/pull/763, we will see how we will manage controlle logs in the future.

PR #677, opened 2022-05-18
[BREAKING] Add multi region support

CalleB3:


CalleB3:
Migration guide

## Governance

### Main.tf

Add to `governance-regional` module:


```terraform
aad_sp_passwords = module.governance_global.aad_sp_passwords
```


## Core

Nothing here is required to do.

## AKS

### Main.tf

In aks-core module, under azad_kube_proxy_config; change to the following:

```terraform
azure_ad_app = module.azad_kube_proxy.data
```

Also, Change `starboard_config` to look like this:

```terraform
starboard_config = {
    client_id                       = module.aks_global.trivy_identity.client_id
    resource_id                     = module.aks_global.trivy_identity.resource_id
    volume_claim_storage_class_name = "managed-csi-zrs"
  }
```

### azad-kube-proxy.tf

Add this file to AKS folder and copy the contents below.


```terraform
locals {
  azad_kube_proxy_url  = "https://aks.${var.dns_zone[0]}"
  azad_kube_proxy_name = "${var.name}-${var.environment}"
}

module "azad_kube_proxy" {
  source = "github.com/xenitab/terraform-modules//modules/azure-ad/azad-kube-proxy?ref=multi_region"

  proxy_url    = local.azad_kube_proxy_url
  display_name = local.azad_kube_proxy_name
  cluster_name = local.azad_kube_proxy_name
}
```


HOW TO ADD ANOTHER CLUSTER IN ANOTHER REGION


## Governance

Once all changes above is applied, we can add another cluster.

### Main.tf

Make a copy of the whole `governance-regional`module, change the name.
inside the module we need to hardcore the region variables:

```terraform
  location                     = "Sweden Central"
  location_short               = "sc"
```


## Core

### Main.tf

Make a copy of the whole `core` module, change the name.

We also need to make another variable for `vnet-config` so that address-spaces in the subnet wont cause any issues.

### ENV.tfvars

Copy the whole declaration of `vnet-config` and make another one with a new name, change to appropriate address-space and CIDR for the vnet and subnet.

### variables.tf

Copy the variable `vnet-config` and make one more, with the same name as above.

## AKS

Make a copy of each module inside AKS, `aks`, `aks-core` and `aks-global`.

modules/azure/aks-global/aad-group.tf
```diff
@@ -1,35 +1,35 @@
...
-  display_name = "${var.aks_group_name_prefix}${var.group_name_separator}${var.subscription_name}${var.group_name_separator}${var.environment}${var.group_name_separator}${each.value.name}${var.group_name_separator}view"
+  display_name = "${var.aks_group_name_prefix}${var.group_name_separator}${var.subscription_name}${var.group_name_separator}${var.environment}${var.group_name_separator}${var.location_short}${var.group_name_separator}${each.value.name}${var.group_name_separator}view"
   #description             = "Members of this group will have view access to the ${each.value.name} namespace."
   prevent_duplicate_names = true
   security_enabled        = true
 }
 
 resource "azuread_group" "edit" {
   for_each     = { for ns in var.namespaces : ns.name => ns }
-  display_name = "${var.aks_group_name_prefix}${var.group_name_separator}${var.subscription_name}${var.group_name_separator}${var.environment}${var.group_name_separator}${each.value.name}${var.group_name_separator}edit"
```
NissesSenap, line 11: What happens in the AAD when this is changed? Is it just the name that gets updated or is it actually a add and removal?
I would assume that we have a number of customers that is dependent on these groups to provide access to there users.
If the name is just updated and the AAD group gid is the same I think this should not create any issues for them.
But if that is not the case we will need to search through the AAD and verify where the current groups are used and help the customers to migrate to the new group.

modules/azure/aks-global/aad-group.tf
```diff
@@ -1,35 +1,35 @@
...
+  display_name = "${var.aks_group_name_prefix}${var.group_name_separator}${var.subscription_name}${var.group_name_separator}${var.environment}${var.group_name_separator}${var.location_short}${var.group_name_separator}${each.value.name}${var.group_name_separator}edit"
   #description             = "Members of this group will have edit access to the ${each.value.name} namespace."
   prevent_duplicate_names = true
   security_enabled        = true
 }
 
 resource "azuread_group" "cluster_admin" {
-  display_name = "${var.aks_group_name_prefix}${var.group_name_separator}${var.subscription_name}${var.group_name_separator}${var.environment}${var.group_name_separator}clusteradmin"
+  display_name = "${var.aks_group_name_prefix}${var.group_name_separator}${var.subscription_name}${var.group_name_separator}${var.environment}${var.group_name_separator}${var.location_short}${var.group_name_separator}clusteradmin"
```
NissesSenap, line 18: As pointed out during our face 2 face review aks-global isn't really global.
Would it be reasonable to create an actual aks-global where we would store a global admin group that could be a part of this group.

That way cluster admins only need to be a part of a single group to get cluster-admin on all the aks clusters.
This might also be a possible good solution for azad-kube-proxy config.

modules/azure/aks-global/variables.tf
```diff
@@ -84,15 +84,3 @@ variable "aks_group_name_prefix" {
...
-  description = "Azure AD Kubernetes Proxy configuration"
-  type = object({
-    cluster_name_prefix = string
-    proxy_url_override  = string
-  })
-  default = {
-    cluster_name_prefix = "aks"
-    proxy_url_override  = ""
-  }
-}
```
NissesSenap, line 98: Might it be worth of actually creating a aks-global and store this config in that module?

NissesSenap (commented):
Great job @jimgus and @CalleB3 !
Added minor comments

modules/azure/aks-global/aad-group.tf
jimgus: I have refactored this part to not change the display name by making it global over regions in order to not bump into any potential problems. Besides, the AD groups does not need to have the regional dimension. Same group can be used in any region.

modules/kubernetes/aks-core/modules.tf
```diff
@@ -504,7 +504,7 @@ module "starboard" {
   cloud_provider                  = "azure"
   client_id                       = var.starboard_config.client_id
   resource_id                     = var.starboard_config.resource_id
-  volume_claim_storage_class_name = "managed-csi-zrs"
+  volume_claim_storage_class_name = var.starboard_config.volume_claim_storage_class_name
```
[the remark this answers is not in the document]
jimgus: Did the same for Prometheus. The sad part is that  it can't be part of the object since it needs a default value, hence it has to be a separate variable :-(

PR #678, opened 2022-05-23
Add k8srequireingressclass together with OPA/library upgrades

NissesSenap:
Update OPA to 3.8.1, gatekeeper-library to 0.12.1 and add k8srequireingressclass constraint.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## Unreleased
 
+### Changed
+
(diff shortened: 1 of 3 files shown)


NissesSenap:
Related to https://github.com/XenitAB/gatekeeper-library/pull/76

NissesSenap:
@phillebaba would it be worth moving https://github.com/XenitAB/terraform-modules/blob/9698261a3684db318d3b6f6b0524e151d9496484/modules/kubernetes/aks-core/modules.tf#L51-L61 and same in EKS in to the OPA module instead?

phillebaba:
Makes sense moving it but I would do it in a different PR.

phillebaba (approved):

PR #679, opened 2022-05-28
Update azurerm provider version

phillebaba:
Adds support to disable run command and update node labels without recreating the node pool.


NissesSenap (approved):
LGTM

PR #689, opened 2022-05-31
Change to calculate AKS resource group instead of using a datasource

phillebaba:
This fixes an issue where the AKS role assignment is recreated when changes to the AKS cluster is made. Its not an optimal solution but at least it does not force the resources to be recreated.


phillebaba:
Closing in favor of #788

PR #691, opened 2022-06-01
Add multi region support

jimgus:
This PR is a refactoring of the Terraform modules to be able to set up multiple AKS clusters in different regions (aka Multi Region) for the same environment. In short, some resources need to be regional (i.e. created once per region) and some resources need global and created only once and will be used by all regions. Changes on a higher level:

xkf-global: New module that contain XKF global Azure AD resources to be used in all regions. This module can be seen as an XKF extension to governance-global and it would have been possible to add these resources to governance-global but that would have made that module too XKF opinionated. This module will be used by both AKS and EKS in the same way as the governance modules and will be added to the governance state.

xkf-global-data:New module.  Some of the resources in xkf-global need to be referenced from other states. We normally do this by using data sources. This module wraps a number of data sources and expose them through the output. This is a more convenient and less error prone way of accessing these resources in contrast to declare those data sources everywhere they need to be accessed.

aks-global: New module that contain AKS global resources. For now it is ACR and the DNS zone. In a multi region setup there shall only be one ACR and one DNS Zone.

aks-regional: This module is renamed from aks-global. Contains AKS common resources per region, i.e. there shall be one instance of this module per region in a Multi Region setup.


NissesSenap (commented):
Initial read through looks good. Added some minor comments.
I will read a few more times, great job @jimgus!

modules/azure/aks-global/variables.tf
```diff
@@ -1,98 +1,67 @@
...
-  type = list(
-    object({
-      name                    = string
-      delegate_resource_group = bool
-    })
-  )
+variable "lock_resource_group" {
+  description = "Lock the resource group for deletion"
+  type        = bool
+  default     = false
```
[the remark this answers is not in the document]
jimgus: We have this option for all other RGs so I added this to be consistent. I think it can be useful for e.g. test setups and ephemeral environments.

modules/kubernetes/aks-core/variables.tf
```diff
@@ -438,6 +449,13 @@ variable "starboard_enabled" {
   default     = true
 }
 
+
+variable "starboard_volume_claim_storage_class_name" {
```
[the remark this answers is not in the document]
jimgus: I will try it out in Sweden Central and if it works I will remove this variable.

modules/azure/aks-regional/trivy.tf
```diff
@@ -0,0 +1,22 @@
+data "azurerm_container_registry" "acr" {
```
NissesSenap, line 1: I think trivy.tf should be able to be in aks-global

modules/azure/aks/azure-metrics.tf
```diff
@@ -4,26 +4,20 @@
...
-}
-
 resource "azurerm_role_assignment" "azure_metrics" {
-  scope                = azurerm_user_assigned_identity.azure_metrics.id
+  scope                = var.azure_metrics_identity.id
   role_definition_name = "Managed Identity Operator"
   principal_id         = var.aks_managed_identity_group_id
 }
 
 resource "azurerm_role_assignment" "azure_metrics_aks_reader" {
```
NissesSenap, line 13: This logic needs to stay in aks thanks to that it is created per aks cluster.

modules/azure/aks-regional/azure-metrics.tf
```diff
@@ -0,0 +1,6 @@
+
+resource "azurerm_user_assigned_identity" "azure_metrics" {
```
NissesSenap, line 2: This should be able to be moved to aks-global

NissesSenap (commented):
I think a number of resources can be moved to aks-global.

PR #692, opened 2022-06-01
Add node ttl to EKS and AKS

phillebaba:
This will add Node TTL to both EKS and AKS. It also sets a TTL of 7 days on all worker nodes. It is however disabled by default so that it can be slowly rolled out.


NissesSenap (dismissed):
LGTM, really looking forward to using this. Great job

NissesSenap (approved):

PR #697, opened 2022-06-03
Set default environment in datadog agent

nissessenap:
Use DD_TAGS on the datadog agent and it adds a tag by default to all events.

This to make it easier for you as a datadog user to always be user that you have a environment taged to your data.

https://github.com/DataDog/datadog-agent/blob/6832bee0ddd0e70bd9f609ef6ca17a058984a563/pkg/config/config_template.yaml#L107-L134

https://docs.datadoghq.com/getting_started/tagging/assigning_tags/?tab=containerizedenvironments#configuration-files


CalleB3 (approved):
LGTM

PR #699, opened 2022-06-07
Update Helm provider to latest version

phillebaba:
Need to upgrade the Helm provider version to the latest to enable OCI chart support.


CalleB3 (approved):
LGTM

PR #718, opened 2022-06-20
Remove create before destroy

phillebaba:
Remove create before destroy on node pools as this has just caused issues when recreating node pools. We should look for an alternative solution to make sure that developers do not remove node pools before adding a new one.


NissesSenap (approved):

NissesSenap:
Should we see this as a breaking change? It's will create new functionality.

phillebaba:
No this just changes the lifecycle of the resource so that it is removed before being created. Nothing breaking.

PR #725, opened 2022-06-28
Temp branch bump azad-kube-proxy to 0.0.34

NissesSenap:
This is pre a big breaking change. If you need it use it.
This should never be merged.


NissesSenap:
We have moved passed this branch now so this shouldn't be an issue any more.

PR #726, opened 2022-06-28
Use git-auth-proxy inside fluxv1

NissesSenap:
This since we have renamed azdo-proxy


NissesSenap:
This won't work, due to that we have moved around the helm chart.
To solve this we would have to do some editing in the index.yaml file to repoint some URI.  For now lets wait with it

PR #729, opened 2022-06-30
override public ip prefix name in aks global

NissesSenap:
This is a workaround to make it possible for us to save a existing ip prefix.
In azure you can move a IP prefix, but you are unable to change the name of it.

To workaround this we have added a override variable, where you can define the name.

Ether before you create your cluster or if you have a existing cluster you need to do a few steps to make use of this variable assuming that you have moved your existing ip prefix to the aks RG.

Add the following to aks.global, assuming that you old prefix is called pip-prefix-dev-we-aks-test-0, for example.
This won't work if you don't have the `-` and a number after.

```.tf
public_ip_prefix_name_override = "pip-prefix-dev-we-aks-test"
```

```shell
# first remove the existing ip prefix
tf state rm 'module.aks_global.azurerm_public_ip_prefix.aks[0]'
# import the new ip prefix
terraform import --var-file ../global.tfvars --var-file ./variables/prod.tfvars --var-file ./variables/common.tfvars 'module.aks_global.azurerm_public_ip_prefix.aks[0]' /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/mygroup1/providers/Microsoft.Network/publicIPPrefixes/pip-prefix-dev-we-aks-test-0
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## Unreleased
 
+### Added
+
+- [#728](https://github.com/XenitAB/terraform-modules/pull/728) Add the possibility to override public ip prefix name in aks global.
+
 ### Changed
 
 - [#714](https://github.com/XenitAB/terraform-modules/pull/714) Ingress-nginx helm chart 4.1.4 and use the chroot functionality.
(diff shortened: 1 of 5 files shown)


NissesSenap:
Have verified that if I don't add the variable this change don't perform any changes to an existing environment.

phillebaba (approved):

PR #739, opened 2022-07-11
Give tenant access to read VPA

NissesSenap:
The tenants needs access to VPA to be able to easily get help with resource suggestions.


jimgus (changes requested):
AWS?

jimgus (approved):

PR #747, opened 2022-07-20
Fix cert-manager webhook AWS config

NissesSenap:
This bug was created in: https://github.com/XenitAB/terraform-modules/pull/733.


NissesSenap:
@jimgus this was probably why you had issues with cert-manager in AWS last time you set it up.

slarwise (approved):

PR #748, opened 2022-07-21
Enable chroot in AWS again

NissesSenap:
In AWS overwrite the default internal-logger-address when having one internal and external ingress defined.
This makes it possible for us to enable chroot again.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Changed
 
 - [#745](https://github.com/XenitAB/terraform-modules/pull/745) Update ingress-nginx to 4.2.0 and disable chroot image in AWS.
+- [#748](https://github.com/XenitAB/terraform-modules/pull/748) Enable chroot image on AWS and set a custom internal-logger-address when running in AWS and multiple internal_load_balancer.
 
 ### Fixed
 
--- a/modules/kubernetes/ingress-nginx/templates/values.yaml.tpl
+++ b/modules/kubernetes/ingress-nginx/templates/values.yaml.tpl
@@ -1,8 +1,6 @@
 controller:
-  %{~ if provider != "aws" ~}
   image:
     chroot: true
-  %{~ endif ~}
 
   replicaCount: 3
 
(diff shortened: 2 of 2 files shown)


NissesSenap:
I have tested that this works after I manually changed the `internal-logger-address`. For more information view: https://github.com/kubernetes/ingress-nginx/issues/8839

We need a upstream PR to add the possibility of setting internal-logger-address through the helm chart.
Will see if I have time for that next week.

When that get's merged we can continue to work on this PR again.

NissesSenap:
Realized that we don't have to do any upstream changes since extraArgs is there. I have verified that this works and it's ready for review.

bittrance (approved):
Frankly, as far as I can understand from https://github.com/kubernetes/ingress-nginx/blob/18ee046b432502aedfed9321f0a9b50bfc009a67/rootfs/Dockerfile-chroot et al, using chroot **reduces** security for the process already running as a container since it creates a special case (note that the default is false) which invites additional bugs. One of the problems with chroot is that it requires constructing a special "environment" (devices, system files, et c) which occasionally needs added files, requiring reevaluation of the entire service. For example, you may want a chroot environment without /etc/passwd. However, later you realize that you need user lookup and want to add the outside /etc/passwd. How do you know whether this is safe?

bittrance:
Further down the rabbit hole, it turns out that ingress-nginx is in fact not using chroot directly, but rather `unshare -R`, see https://github.com/kubernetes/ingress-nginx/blob/18ee046b432502aedfed9321f0a9b50bfc009a67/rootfs/nginx-chroot-wrapper.sh#L18 and https://github.com/util-linux/util-linux/blob/60803291acc7f431fd7750de7b2a2f1d45db1f4d/sys-utils/unshare.c#L1071 . This means that before calling `chroot(2)` it calls `unshare(2)`. Unfortunately, the Ingress controller does not in fact pass any unshare flags, meaning that it does not disassociate any namespaces (such as mount, user or ipc namespaces). Still, it makes the setup a bit less hacky. Hopefully, the controller will over time unshare more of its execution context so that less sensitive information is exposed.

This has no immediate bearing on this PR, but the use of unshare indicates that the complexity of `Dockerfile-chroot` will retain the maintainers' attention and will hopefully remain secure and the the ingress nginx will eventually get a reduced surface area.

PR #749, opened 2022-07-21
Update gatekeeper to 3.9.0

NissesSenap:
Update to latest gatekeeper-lib.
These settings include support for debug containers in k8s 1.23.


NissesSenap:
I have verified it and the upgrade works as intended.

dachrillz (approved):

PR #750, opened 2022-07-22
Run helm chart in client_only mode to fix linkerd CRD install issues.

NissesSenap:
The linkerd helm chart have defined which versions of k8s it supports.
Sadly the helm provider have hardcoded API k8s client version that it uses to apply the config. Thanks to this we are unable to apply the latest linkerd config through terraform.

This is a try to workaround this issue.

### DIFF
--- a/modules/kubernetes/helm-crd/main.tf
+++ b/modules/kubernetes/helm-crd/main.tf
@@ -19,6 +19,7 @@ data "helm_template" "this" {
   name         = var.chart_name
   version      = var.chart_version
   include_crds = true
+  validate     = true
   api_versions = ["apiextensions.k8s.io/v1/CustomResourceDefinition"]
 
   dynamic "set" {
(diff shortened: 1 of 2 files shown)


NissesSenap:
Out of my depth, @phillebaba spoke about clientOnly but reading the code it's not possible to set this value. Probably is but I need to do some special thing in terraform.

https://github.com/hashicorp/terraform-provider-helm/blob/c97454efc586eba8b1170364ce160b8729cfe6a2/helm/data_template.go#L430-L433
There is no d.Get and clientOnly is hard coded.

NissesSenap:
And it was rather simple I looked in the wrong resource....
To early or something.

NissesSenap:
Going to try if this is a potential workaround: https://github.com/XenitAB/terraform-modules/pull/751

NissesSenap:
We solved this in https://github.com/XenitAB/terraform-modules/pull/768

PR #755, opened 2022-07-25
Tmp remove aks max surge

NissesSenap:
Getting this error when trying to create a spot instance:
╷
│ Error: creating Node Pool: (Agent Pool Name "memoryspot1" / Managed Cluster Name "node1" / Resource Group "rg1"): containerservice.AgentPoolsClient#CreateOrUpdate: Failure sending request: StatusCode=400 -- Original Error: Code="InvalidParameter" Message="The value of parameter agentPoolProfile.upgrade.maxSurge is invalid. Error details: Spot pools can't set max surge. Please see https://aka.ms/aks-naming-rules for more details." Target="agentPoolProfile.upgrade.maxSurge"
│ 
│   with module.aks1.azurerm_kubernetes_cluster_node_pool.this["memoryspot1"],
│   on .terraform/modules/aks1/modules/azure/aks/aks.tf line 65, in resource "azurerm_kubernetes_cluster_node_pool" "this":
│   65: resource "azurerm_kubernetes_cluster_node_pool" "this" {
│


NissesSenap:
We stopped using memory spot instances for now.
If this becomes and issue in the future it should be straight forward to fix.

PR #763, opened 2022-08-23
AKS audit logs for aks

NissesSenap:


NissesSenap:
And computer says no:

```
Service returned an error. Status=400 Code="BadRequest" Message="Resources should be in the same region.
```

So we will aim to have this in core instead.

PR #768, opened 2022-08-24
Use OCI helm chart in xenit linkerd fork

NissesSenap:
In the linkerd fork we have removed the helm chart k8s version limitation.
We also provide the helm chart through OCI.

It uses the helm-crd-oci that is fixed in: https://github.com/XenitAB/terraform-modules/pull/767

### DIFF
--- a/modules/kubernetes/linkerd/main.tf
+++ b/modules/kubernetes/linkerd/main.tf
@@ -192,10 +192,11 @@ resource "kubernetes_secret" "webhook_issuer_tls" {
 # Install linkerd-cni helm chart
 #tf-latest-version:ignore
 resource "helm_release" "linkerd_cni" {
-  chart       = "oci://ghcr.io/xenitab/helm-charts/linkerd2-cni"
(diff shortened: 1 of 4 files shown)


modules/kubernetes/linkerd/main.tf
```diff
@@ -193,10 +193,10 @@ resource "kubernetes_secret" "webhook_issuer_tls" {
 #tf-latest-version:ignore
 resource "helm_release" "linkerd_cni" {
   repository  = "https://helm.linkerd.io/stable"
-  chart       = "linkerd2-cni"
+  chart       = "linkerd-control-plane"
   name        = "linkerd-cni"
   namespace   = kubernetes_namespace.cni.metadata[0].name
-  version     = "2.11.3"
+  version     = "30.3.0"
```
NissesSenap, line 199: Note the versions follows some semver of the helm chart and not the app version any more. 30.3.0 is 2.12.0

CalleB3 (approved):

PR #775, opened 2022-09-02
Update Falco helm chart to 2.0.16

NissesSenap:
The 2.0 chart contains multiple breaking changes.

Manual work is needed due to updates in DS label selector.

```
│ Error: cannot patch "falco" with kind DaemonSet: DaemonSet.apps "falco" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app.kubernetes.io/instance":"falco", "app.kubernetes.io/name":"falco"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable
│ 
│   with module.aks1_core.module.falco["falco"].helm_release.falco,
│   on .terraform/modules/aks1_core/modules/kubernetes/falco/main.tf line 34, in resource "helm_release" "falco":
│   34: resource "helm_release" "falco" {
```

Run

```shell
k delete ds falco -n falco
```

And then update the falco module.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 - [#778](https://github.com/XenitAB/terraform-modules/pull/778) Upgrade AWS calico to 3.24.
 - [#776](https://github.com/XenitAB/terraform-modules/pull/776) [Breaking] Remove default value for unique_suffix in Azure core module.
+- [#775](https://github.com/XenitAB/terraform-modules/pull/775) Update falco helm chart to 2.0.16.
 
 ## 2022.08.3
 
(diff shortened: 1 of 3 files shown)


NissesSenap:
Verified that this works in AKS. I need to verify the same thing on EKS.
But I don't want to downgrade my existing test cluster and there is an issues with my node version. When https://github.com/falcosecurity/test-infra/pull/824 gets merged and the images are built my cluster should start working again I will be able to verify it on EKS.

NissesSenap:
Instead of doing the manual work should we temporary add
[force_update](https://registry.terraform.io/providers/hashicorp/helm/latest/docs/resources/release#force_update) to the falco helm chart. This way it should automatically fix the upgrade without manual work.

jimgus (approved):

jimgus (dismissed):

NissesSenap:
It didn't help to add the force_update flag.
So I will remove it.

jimgus (approved):

NissesSenap:
This is breaking from a k8s point of view since we can't update an already existing DS with new selectors.
But from a tf point of view nothing is breaking.

But sadly you need to follow the instructions about and manually remove the ds before running apply.

PR #776, opened 2022-09-02
Remove default value on unique_suffix in core module

NissesSenap:
It's really easy to make configuration mistakes thanks to this default value.
For people not defining it you will need to set

```
unique_suffix = ""
```


jimgus (approved):
Very good!

Issue #728, opened 2022-06-30
fluxv2 upgrade to post 0.28.0


NissesSenap:
Fluxv2 have breaking changes in 0.28.0

https://github.com/fluxcd/flux2/releases/tag/v0.28.0

We need to go through them, upgrade our clusters and make a smooth transition to the latest version for our tenants and our self.

PR #777, opened 2022-09-02
Upgrade flux v2 to 0.17.0 and update source CR to v1beta2

nissessenap:
Fix https://github.com/XenitAB/terraform-modules/issues/728


NissesSenap:
It's breaking from a user point of view, if they manage any source objects outside of terraform.
But if you only use the git resource that we generate in this module you won't need to do a thing. 
Look at the docs for more information: https://github.com/XenitAB/xenitab.github.io/pull/152

PR #784, opened 2022-09-07
Upgrade k8s to 1.24

NissesSenap:
Update your aks/main.tf to support the new version of 1.24.

```
provider "kubernetes" {
  alias                  = "aks1"
  host                   = module.aks1.kube_config.host
  cluster_ca_certificate = module.aks1.kube_config.cluster_ca_certificate
  exec {
    api_version = module.aks1.kube_config.api_version
    args = module.aks1.kube_config.args
    command = module.aks1.kube_config.command
  }
}

provider "helm" {
  alias = "aks1"
  kubernetes {
    host                   = module.aks1.kube_config.host
    cluster_ca_certificate = module.aks1.kube_config.cluster_ca_certificate
    exec {
      api_version = module.aks1.kube_config.api_version
      args = module.aks1.kube_config.args
      command = module.aks1.kube_config.command
    }
  }
}

provider "kubectl" {
  alias                  = "aks1"
  host                   = module.aks1.kube_config.host
  cluster_ca_certificate = module.aks1.kube_config.cluster_ca_certificate
  exec {
    api_version = module.aks1.kube_config.api_version
    args = module.aks1.kube_config.args
    command = module.aks1.kube_config.command
  }
  load_config_file       = false
}
```

PR #786, opened 2022-09-08
Add support for kubernetes version 1.24 for AKS and k8s tf login through a local file

NissesSenap:


NissesSenap:
Lets see if this works better then: https://github.com/XenitAB/terraform-modules/pull/784/

NissesSenap:
This will never work since the file needs to be created before the plan stage.
Since this never happens we have to come up with something else.

PR #792, opened 2022-09-09
Migrate public ingress-nginx to seperate ns

NissesSenap:


modules/kubernetes/prometheus/charts/prometheus-extras/templates/monitors.yaml
```diff
@@ -64,7 +64,7 @@ metadata:
 spec:
   endpoints:
   - interval: 30s
-    port: metrics
+    port: http-metrics
```
NissesSenap, line 67: We need to monitor the private ns as well. 
Need to add a variable.

NissesSenap:
This was solved in: https://github.com/XenitAB/terraform-modules/pull/793

PR #798, opened 2022-09-14
WIP: Implemenet Cilium as CNI

CalleB3:
Add cilium CNI

Ignore azure-container-configured-network-policy
AKS tf don't understand this since we are using BYOCNI

Notes:
We appear to be needing to use the default ServiceMonitors.
On the other hand we could create the services by ourselves aswell.


NissesSenap:
This PR is currently on hold. But we will keep it open.
Changing it to WIP

landerss1:
This PR is superceeded by [#1142](https://github.com/XenitAB/terraform-modules/pull/1142).

PR #807, opened 2022-09-20
Increase nginx ingress min availible from one to two

phillebaba:
The default value for the nginx PDB is 1, which I do not like. Would rather have it to be 2 so that only one instance is replaced at a time.


NissesSenap (approved):
I don't think this will create any issues in AWS. The switch of ingress-nginx will just go slower due to they normally scale up new nodes for it.

PR #808, opened 2022-09-21
Increase prometheus default disk size to 15Gi

NissesSenap:


NissesSenap:
We need default values in tf objects before this is a good option.

PR #815, opened 2022-09-23
Make datadog tolerate all node taints

phillebaba:
Datadog DS agents should run on all nodes even if tainted.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - [#810](https://github.com/XenitAB/terraform-modules/pull/810) Upgrade terraform to 1.3.0.
 - [#810](https://github.com/XenitAB/terraform-modules/pull/810) Update provider versions.
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


NissesSenap:
Hmm when looking at Cilium we are currently thinking about adding the following to AWS.

```
/etc/eks/bootstrap.sh ${cluster_name} --kubelet-extra-args '--register-with-taints=node.cilium.io/agent-not-ready=true:NoExecute' '--node-labels=eks.amazonaws.com/capacityType=ON_DEMAND,eks.amazonaws.com/nodegroup=${node_group}' --b64-cluster-ca $B64_CLUSTER_CA --apiserver-endpoint $API_SERVER_URL --use-max-pods false
```

So when the kubelet registers the node it should disable any pods except the cilium DS.
When the cilium DS starts it will remove this taint and allow other workloads on the node.

If you set

```
      tolerations:
        - operator: Exists
```

won't this ignore that taint and make it start.
I can see a similar issue when we are using linkerd and the datadog pod would start before linkerd is up and running.

In short should we make the taint a bit more restricted?

NissesSenap (approved):

phillebaba:
Lets take the discussion when Cilium is merged. I think that there are a lot of daemonsets which need to be adjusted in that case.

NissesSenap (approved):

PR #821, opened 2022-09-29
Default ingressClassName for ingress_healthz

NissesSenap:
This to remove issues in e2e tests.

When creating a vanilla setup I got the following error.

```
│ Error: admission webhook "validation.gatekeeper.sh" denied the request: [k8srequireingressclass] Ingress class name has to be set
│
│   with module.aks1_core.module.ingress_healthz["ingress-healthz"].helm_release.ingress_healthz,
│   on .terraform/modules/aks1_core/modules/kubernetes/ingress-healthz/main.tf line 34, in resource "helm_release" "ingress_healthz":
│   34: resource "helm_release" "ingress_healthz" {
│
```

this is probably due to OPA not beeing up and running.

We can solve this problem with a workaround like this or we can try to come up with a way of getting OPA to actually return true when things is up and running.
If I run plan & apply again OPA does the job as intended.


modules/kubernetes/aks-core/modules.tf
```diff
@@ -238,10 +238,11 @@ module "ingress_healthz" {
...
 
-  environment     = var.environment
-  dns_zone        = var.cert_manager_config.dns_zone[0]
-  location_short  = var.location_short
-  linkerd_enabled = var.linkerd_enabled
+  environment            = var.environment
+  dns_zone               = var.cert_manager_config.dns_zone[0]
+  location_short         = var.location_short
+  linkerd_enabled        = var.linkerd_enabled
+  public_private_enabled = var.ingress_config.public_private_enabled
```
NissesSenap, line 245: Do we think it's a good idea to use this config variable that should be used for ingress-nginx?

modules/kubernetes/ingress-healthz/main.tf
```diff
@@ -21,6 +21,10 @@ terraform {
   }
 }
 
+locals {
+  ingress_class_name = var.public_private_enabled == true ? "nginx" : "nginx-public"
```
jimgus, line 25: The other way around

jimgus (changes requested):

jimgus (approved):
LGTM!

PR #831, opened 2022-10-19
Add conditional if tenant doesnt use unique_suffix

CalleB3:


NissesSenap:
This have been solved other ways.

PR #840, opened 2022-10-27
Add support for ARM VMs in AKS

phillebaba:
This change makes it possible to use ARM VMs in AKS. It should probably not be used initially in production but gives the option to begin experimenting.


jimgus (approved):
Nice!

PR #841, opened 2022-10-27
Change ACR SKU to Standard

phillebaba:
This sets all ACR SKUs to Premium. I do not see a point in running a Basic SKU as Premium includes a lot more storage. Plus the increased bandwidth does not hurt.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - [#837](https://github.com/XenitAB/terraform-modules/pull/837) Update TFLint to 0.42.
 - [#838](https://github.com/XenitAB/terraform-modules/pull/838) Update git-auth-proxy to v0.8.1.
 - [#840](https://github.com/XenitAB/terraform-modules/pull/840) Add support for ARM VMs in AKS.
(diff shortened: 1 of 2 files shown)


NissesSenap:
Will this recreate the acr or not when upgrading?
How much is the process difference?
I think it's a good idea, this provides oci compatibility as well so we can store helm charts and other fun stuff in acr.

jimgus:
Agree with above comment, it would be good to know if there are any upgrade impacts or not

phillebaba:
Sorry this was actually meant to be Standard and not Premium. Changing the SKU has no affect more than it just changing.

**Basic**
> A cost-optimized entry point for developers learning about Azure Container Registry. Basic registries have the same programmatic capabilities as Standard and Premium (such as Azure Active Directory authentication integration, image deletion, and webhooks). However, the included storage and image throughput are most appropriate for lower usage scenarios.

**Standard**
> Standard registries offer the same capabilities as Basic, with increased included storage and image throughput. Standard registries should satisfy the needs of most production scenarios.


https://learn.microsoft.com/en-us/azure/container-registry/container-registry-skus

Reading this it is reasonable that we should be using Standard and not Basic.

NissesSenap:
Looking at the pricing it seems like a no brainer as well.
https://azure.microsoft.com/en-us/pricing/details/container-registry/

NissesSenap (approved):

PR #842, opened 2022-10-27
Refactor Prometheus variables to make use of optionals

phillebaba:
This PR refactors Prometheus configuration and makes use of optional. This will hopefully reduce the need to repeat default values that risks making them wrong. It also removes some variables that are no longer needed that just increase the complexity.


jimgus:
Make sure to update https://github.com/XenitAB/xkf-templates accordingly

landerss1:
Checking against current configurations for our tenants, I can't see any that use these defaults, they are all different, so this PR will not add much value except for maybe new clusters. Closing this issue.

PR #846, opened 2022-11-02
Add XKF prefix to node pool label

phillebaba:
This changes the default label added to nodes indicating the node pool it is part of. The reason is to make it obvious that it is a label set by XKF. This change could be considered breaking if people have node selectors for the `node-pool` label, but as far as I know it should not be an issue.


NissesSenap (approved):
It's nice to be consistent.
LGTM

PR #854, opened 2022-11-14
Explicitly set Prometheus version that should be used

phillebaba:
Currently the Prometheus Operator will set the image tag to latest, which works but has its issues. One is that it is not clear which Prometheus version is being used. The other is that latest could be an old version with a backported fix. Setting the version explicitly will make sure that a specific version is being used. It will add the burden to update the version but I think it is worth it.


NissesSenap (approved):
Agreed

PR #856, opened 2022-11-14
Update falco to v0.33.0 and falco-exporter v0.8.0

NissesSenap:
Also change from `falco.grpcOutput.enabled` to `falco.grpc_output.enabled`.
This to use the new configmap injection that the helm chart uses now a days.


NissesSenap:
Finally it is done!

jimgus (approved):

PR #858, opened 2022-11-15
Use AKS NAT gateway

NissesSenap:


NissesSenap:
For now it seems like NAT gateway can't be made zone redundant.
Instead we have to rewrite our network stack to use 3 subnets, 3 public ip prefixes and 3 nat gateways.
It would increase the cost by allot and it also creates allot more work then we appreciated.

So for now we will close this PR and see how we prioritize it against this new information.

PR #862, opened 2022-11-17
Simplify azad kube proxy locals

phillebaba:


landerss1:
This PR is seriously outdated. We don't use locals in this module anymore. Closing.

PR #867, opened 2022-11-18
Fix DNS Zones variable naming

phillebaba:
This change fixes variable naming to be plural when dns zones are a list.


modules/kubernetes/eks-core/modules.tf
```diff
@@ -15,16 +15,16 @@ locals {
     "velero",
     "promtail"
   ]
-  dns_zone = {
+  route53_zones = {
```
NissesSenap, line 18: Hmm why don't you want to use dns zones any more?
From my point of view we are working with the construct of dns zones, not a specific product.
I can live with it but just a thought.

NissesSenap (approved):
Minor comment, I understand if you don't agree.
Great to get this consistent :)

modules/kubernetes/eks-core/modules.tf
phillebaba: Because this local variable does not represent dns_zones as in all other uses. Its an iteration of Route53 zone ids. Having the same name would confuse more than it helps. Giving it a different name makes it clear that it contains other data.

landerss1:
This PR is seriously outdated. We don't support EKS anymore. Closing.

PR #869, opened 2022-11-24
Add support to set taint on EKS node pool

NissesSenap:


phillebaba (approved):
Looks good, in the future I would suggest that we add validation for the values.

jimgus (approved):

NissesSenap:
I couldn't come up with a nice way of doing for loops in for loop in validation.
I would love to get some help on this one if you have a good idea how to write it.

PR #871, opened 2022-11-24
Change ingress-nginx affinity match labels

jimgus:
In AWS we have seen that ingress-nginx pods might not be scheduled schedule across all 3 AZs. The implication to that is that the NLB in the zone that have no ingress-nginx pod scheduled is taken out of service. This PR changes that the matchExpressions for private and public are not the same since they should be scheduled separately (bug). The other change is that the PriorityClass is set to High to allow for preemption to make sure all zones gets a pod scheduled.


phillebaba (approved):
LGTM

NissesSenap (approved):

PR #877, opened 2022-12-01
Update Kube Prometheus Stack version

phillebaba:
Contains a change which is breaking but solved with a force update in the helm release. This change will recreate the node exporter daemon set when updated.

 https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack#from-39x-to-40x

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Changed
 
 - [#879](https://github.com/XenitAB/terraform-modules/pull/874) Update Promtail Helm chart to 6.6.2.
+- [#877](https://github.com/XenitAB/terraform-modules/pull/877) Update Kube Prometheus Stack to 42.1.1.
 
 ## 2022.12.1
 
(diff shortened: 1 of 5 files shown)


jimgus (commented):
Don't know if "If you use your own custom ServiceMonitor or PodMonitor, please ensure to upgrade their selector fields accordingly to the new labels" has any implications on us?

NissesSenap:
You also need to update the aks_core and eks_core CRD:s
https://github.com/XenitAB/terraform-modules/blob/bb2b63445b328df2a1ef19f695da2a20ded926a5/modules/kubernetes/aks-core/modules.tf#L471-L478

phillebaba:
@jimgus that comment was weird but it refers to if you have custom service monitors on the node exporters, you need to update the selectors. As far as I understand we do not create a service monitor for node exporter on our own.

jimgus (approved):

NissesSenap:
Even if the helm_force flag was set we still get DS issues.

│ Error: failed to replace object: DaemonSet.apps "prometheus-prometheus-node-exporter" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app.kubernetes.io/instance":"prometheus", "app.kubernetes.io/name":"prometheus-node-exporter"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable
│ 
│   with module.aks1_core.module.prometheus["prometheus"].helm_release.prometheus,
│   on .terraform/modules/aks1_core/modules/kubernetes/prometheus/main.tf line 34, in resource "helm_release" "prometheus":
│   34: resource "helm_release" "prometheus" {

PR #884, opened 2022-12-06
get-nodes have to be a clusterrolebinding

NissesSenap:
Since nodes are a cluster resource they also need to have a clusterrolebinding.
This was solved in #661 for aks


phillebaba (approved):

modules/kubernetes/eks-core/k8s-cluster-role-binding.tf
```diff
@@ -37,3 +37,24 @@ resource "kubernetes_cluster_role_binding" "cluster_view" {
     api_group = "rbac.authorization.k8s.io"
   }
 }
+
+resource "kubernetes_cluster_role_binding" "get_node" {
```
CalleB3, line 41: Maybe we want to use the same naming as aks. e.g `get_nodes`?

NissesSenap: Fixed

CalleB3 (dismissed):

CalleB3 (approved):

PR #893, opened 2022-12-15
Put labels and taints in AWS template file

NissesSenap:


NissesSenap:
This shouldn't be needed, we just need to tag the ASG.

PR #899, opened 2022-12-28
Remove instance label from API Server metrics

jimgus:
The instance label contain two different values on AWS and hence duplicates the number of time series


simongottschlag (approved):
lgtm :^)

PR #914, opened 2023-01-13
Set EKS max unavailible percentage to same value as AKS

phillebaba:
This will speed up node pool deletions by allowing parallel changes to nodes.


phillebaba:
I may have missunderstood this setting. We should explore this setting a bit more.
https://docs.aws.amazon.com/eks/latest/APIReference/API_NodegroupUpdateConfig.html

landerss1:
Won't do. AWS is no longer supported.

PR #915, opened 2023-01-13
Opa gatekeeper pdb constraint

NissesSenap:
Use the new PDB rule to make sure that a user can't define a PDB minAvliable bigger then an HPA/deployment number.


NissesSenap:
I have hit times when OPA admission webhook don't protect against miss configuration:
For more info see:
https://open-policy-agent.github.io/gatekeeper/website/docs/failing-closed/

NissesSenap:
It seems like the admisionwebhook isn't perfect, I have created an issue upstream about this:
https://github.com/open-policy-agent/gatekeeper-library/issues/279

Lets see if we get any luck

landerss1:
With Azure Policy, we would have to do this in the azure-policy module in a different way.

landerss1:
Closing this as this wasn't solved upstream.

PR #917, opened 2023-01-20
Remove datasource for azuread_groups in xkf_governance_global

CalleB3:
Saw issues with 1 customer that tags that were added outside of terraform, then removed by terraform contributed to resources being forced to be replaced when using the datasource prior to this change. This change will simply output the values from one module to another instead of the datasource.

Also removed the variable `azure_ad_group_prefix` since its not being used.

Tested and works.
Also tested with a customer that doenst have any issues and got:
`No changes. Your infrastructure matches the configuration.`


simongottschlag:
@CalleB3 please add a PR description with the problem being solved with this change.

simongottschlag (approved):
LGTM! :^)

NissesSenap (commented):
Are we not missing output in governance-regional to feed xkf-governance-global with?
If you can prepare a PR in the templates repo as well that would be great.

NissesSenap:
Ahh I read your comment now. So your thought is to get the data directly from governance-global instead.
I will read a bit more code and come back to you

NissesSenap (approved):
LGTM
I read trough the changes and the modules together with @jimgus .
We think that we also can remove the depends_on normally used in xkf_governance_global

```
module "xkf_governance_global" {
  source = "github.com/xenitab/terraform-modules//modules/azure/xkf-governance-global?ref=2023.01.2"
  depends_on = [
    module.governance_regional
  ]
``` 

Got nothing to do with this PR but a good thing to take care of in the template PR.

PR #921, opened 2023-01-31
Monitor if logs is missing

NissesSenap:
The log storage account should constantly get audit logs sent to it.
If not we should notice it.

### DIFF
--- a/modules/azure/core/variables.tf
+++ b/modules/azure/core/variables.tf
@@ -97,3 +97,14 @@ variable "unique_suffix" {
   description = "Unique suffix that is used in globally unique resources names"
   type        = string
 }
+
+variable "notification_email" {
+  description = "Email address to send alert notifications"
+  type        = string
+}
+
+variable "alerts_enabled" {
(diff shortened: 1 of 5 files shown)


modules/azure/core/variables.tf
```diff
@@ -97,3 +97,14 @@ variable "unique_suffix" {
   description = "Unique suffix that is used in globally unique resources names"
   type        = string
 }
+
+variable "notification_email" {
+  description = "Email address to send alert notifications"
+  type        = string
+}
```
NissesSenap, line 104: Since the notification_email don't have any default value we have to define it.
Not a problem in our templates repo though.

We can of course add it and set it to something like `""` but if we do then we might make a human error and not sending the errors to anything. 

We could also add our own email in here but we haven't done so in other locations so I guess we are trying not to spread it to much.

jimgus (approved):

bittrance (approved):
I'm not familiar with dynamic criteria, but otherwise looks reasonable.

PR #934, opened 2023-02-13
Add certificate permissions for resource group AAD group

CalleB3:
This is to fix access issues for some customers that has started to utilize certificates in Azure KeyVault


NissesSenap (approved):

NissesSenap:
Since the pods inside the tenant namespace should be able to access the certificates this makes sense.
LGTM

PR #936, opened 2023-02-13
Add Spegel to AKS and EKS

phillebaba:
This change adds Spegel to both EKS and AKS by default.


modules/kubernetes/spegel/main.tf
```diff
@@ -0,0 +1,38 @@
...
+  metadata {
+    name = "sepgel"
+    labels = {
+      name                = "spegel"
+      "xkf.xenit.io/kind" = "platform"
+    }
+  }
+}
+
+resource "helm_release" "this" {
```
NissesSenap, line 32: Is there any known differences between AKS/EKS?

NissesSenap (approved):

PR #944, opened 2023-02-22
[Breaking] Upgrade grafana-agent to 0.2.12

NissesSenap:
The deployment of the operator updates a few selectors....

```
│ Error: cannot patch "grafana-agent-operator" with kind Deployment: Deployment.apps "grafana-agent-operator" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app.kubernetes.io/instance":"grafana-agent-operator", "app.kubernetes.io/name":"grafana-agent-operator"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable
```

You manually need to deplete the deployment before applying the change.

```
k delete deployments.apps grafana-agent-operator -n grafana-agent
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## Unreleased
 
+### Changed
+
+- [#944](https://github.com/XenitAB/terraform-modules/pull/944) Upgrade grafana-agent to 0.6.0.
+
 ## 2023.02.3
 
 ### Changed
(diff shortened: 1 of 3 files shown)


landerss1:
Grafana agent has been deprecated and will be replaced with Grafana Alloy. Upon migration to install with Flux, we understood that upgrading the Grafana agent to latest wasn't something that could be easily done. Furthermore, we are re-thinking the way we handle monitoring, so it wouldn't be well spent time trying to upgrade this module at this point. Therefore, closing this.

PR #945, opened 2023-02-28
Set flux git repo fetch to 3m and 120s timeout

NissesSenap:


CalleB3:
LGMT, no sure how it would effect customers though

NissesSenap:
It will increase the default delay for the customer from 1 min to 3 min per reconciliation.
So in short it takes some more time to deploy changes after merging.

phillebaba (approved):
Looks good to me, we should re-evaluate in the future after updating Flux.

PR #946, opened 2023-03-01
Enable configuration for private and public ingress controllers

phillebaba:


modules/kubernetes/ingress-nginx/variables.tf
```diff
@@ -27,38 +15,53 @@ variable "default_certificate" {
...
+variable "config" {
+  type = object({
+    allow_snippet_annotations = optional(bool, false)
+    http_snippet              = optional(string, "")
+    extra_config              = optional(map(string), {})
+    extra_headers             = optional(map(string), {})
+  })
+}
+
+variable "config_public" {
```
phillebaba, line 39: `config_public` is not a great name. Does anyone have a better idea for the name?

bittrance: Not sure, but maybe `private_ingress_customization` and `public_ingress_customization`?

bittrance: ... or `extra_{private,public}_customization` given that "extra" is a thing in this context?

phillebaba: Customization may be a good word. The issue comes from the fact that we went from single attributes to duplicating them in multiple objects to enable global and override in a sane manner. So `config` will be used if `private-public` is false and also as a global default if `private-public` is true. Then the other two blocks enable specific overrides for the public and private ingress controllers.

NissesSenap (approved):
LGTM, great job and great future proofing with the new name ingress_nginx_config

phillebaba:
This change is sadly breaking to the configuration as it requires manual configuration changes. Nothing much but will require updating some variables.

Issue #950, opened 2023-03-06
Core broken due to subnet_private_endpoint_network_policies_enabled


NissesSenap:
This issue was created in https://github.com/XenitAB/terraform-modules/pull/943

```shell
│ Error: Conflicting configuration arguments
│ 
│   with module.core.azurerm_subnet.this["sn-dev-we-core-servers"],
│   on .terraform/modules/core/modules/azure/core/subnets.tf line 13, in resource "azurerm_subnet" "this":
│   13:   enforce_private_link_endpoint_network_policies = local.subnet_private_endpoints[each.value.subnet_short_name]
│ 
│ "enforce_private_link_endpoint_network_policies": conflicts with private_endpoint_network_policies_enabled
```

It looks like there is a bug in the provider:

https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/subnet#private_endpoint_network_policies_enabled

Even if I disable `private_endpoint_network_policies_enabled` it still complains. We could probably solve this with a dynamic resources as well.
But I will first see if there is any issues open around it.

NissesSenap:
FYI @linus-lysell

CalleB3:
This was fixed in #952

PR #943, opened 2023-02-22
Fixes #942: Core/azurerm_subnet: Make private_endpoint_network_policies_enabled configurable

linus-lysell:
Core/azurerm_subnet: Make "private_endpoint_network_policies_enabled" configurable

PR #952, opened 2023-03-06
Fix subnet issue from #943

NissesSenap:
- Use the existing for loop logic
- Remove var.subnet_private_endpoint_network_policies_enabled
- Change core to use new private_endpoint_network_policies_enabled in core subnet config
- Solves #950


phillebaba (dismissed):

linus-lysell (approved):

phillebaba (dismissed):

modules/azure/core/variables.tf
```diff
@@ -24,30 +24,21 @@ variable "vnet_config" {
...
-      cidr                                      = string
-      service_endpoints                         = list(string)
-      aks_subnet                                = bool
-      private_endpoint_network_policies_enabled = optional(bool, true)
+      name                                          = string
+      cidr                                          = string
+      service_endpoints                             = list(string)
+      aks_subnet                                    = bool
+      private_link_service_network_policies_enabled = optional(bool, true)
+      private_endpoint_network_policies_enabled     = optional(bool, true)
```
NissesSenap, line 32: So at least in AKS these values are already set. And my understanding is that they are also set in the portal by default and in v4 of azurerm the will be true.
So in short true is probably the way to go.

phillebaba: My opinion is that unless it needs to be configured it should not have a variable. I think it would be better to set these to true in the module instead.

phillebaba:
I have changed the configuration to just use hard coded values. It is what is default in the Azure UI and Azure CLI, so I think it's a good choice for us also.

CalleB3 (approved):
LGTM

Issue #941, opened 2023-02-17
Go over to trivy-server in the trivy-operator


NissesSenap:
When https://github.com/aquasecurity/trivy-operator/pull/979 gets merged we should be able to go over to trivy-server from the trivy-operator. Thus we can remove one helm chart from our terraform solution.

NissesSenap:
Also waiting for https://github.com/aquasecurity/trivy-operator/pull/982

PR #953, opened 2023-03-06
Switch to new trivy operator and trivy server

CalleB3:


modules/kubernetes/trivy/templates/trivy-operator-values.yaml.tpl
```diff
@@ -10,7 +10,17 @@ trivy:
   mode: ClientServer
   severity: MEDIUM,HIGH,CRITICAL
   ignoreUnfixed: true
-  serverURL: "http://trivy.trivy.svc.cluster.local:4954"
+  %{~ if provider == "aws" ~}
+  serviceAccount:
+    annotations:
+      eks.amazonaws.com/role-arn: ${trivy_operator_role_arn}
```
NissesSenap, line 16: Hmm isn't there a service account created for the trivy server?
You will probably need to use var.trivy_role_arn this because the role that gets created is specific to the serviceAccount.
You might also need to update the role_arn to match the new serviceAccount that gets created by the oeprator.

NissesSenap (commented):
I think it looks good, but I think you need to take an extra look on the EKS side.
See comment

modules/kubernetes/trivy/templates/trivy-operator-values.yaml.tpl
CalleB3: From what i can tell, trivy-server uses the SA created for trivy-operator.
What do you mean with "update the role_arn to match the new serviceAccount that gets created by the oeprator." Which role is this?

CalleB3: Currently there are 2 arn's,`trivy_operator_role_arn`  and `trivy_role_arn`.
There should be no use for both?

NissesSenap: Ahh okay if there is only one SA, you can remove the one called `trivy_role_arn`  just as you say.
https://github.com/XenitAB/terraform-modules/blob/1aff96b7e23bb6038b89582c84eab990cc21dbcf/modules/aws/eks/iam.tf#L248

Good job

CalleB3:
Fixes #941

landerss1:
Support for EKS has been deprecated. Furthermore, we are already running the trivy-operator. Closing this.

Issue #501, opened 2021-12-28
Support resource naming override


tvuorenmaa89:
### Description

When using the core module to deploy resources there we have identified a need to override the standard naming of resources. 

**Scenario 1 - Subscription short name** 

When using the core-module to deploy multiple core setups in customers environments we will end up with multiple resources with the same name. This is because we use the core-module to deploy any kind of Landing zones in Azure.

**Example:** 
Using the core module multiple times will result in multiple virtual networks called vnet-prod-we-core.

Suggestion: 
Would it be possible to introduce subscription short as a optional attribute in the naming? 
**vnet-prod-we-lzvw-core** where lzvw is shorten for Landing zone Virtual Workspace. 

**Scenario 2 - CAF Naming Convention / Customer specific**

Microsoft have a naming convention in Azure which we try to align to but it does not follow a strict pattern: 

**Example:**
Subnet - 'snet-subscription-region-###'
Public  IP - pip-"vm name or app name"-environment-region-###
Route Table - route-"route table name"

More examples: https://docs.microsoft.com/en-us/azure/cloud-adoption-framework/ready/azure-best-practices/resource-naming

In this case it would be useful to be able to override the naming convention to align with CAF. 
The same scenario goes for deployments where customer have a established naming convention we need to follow for new resources.

simongottschlag:
@ahockersten can you make sure we get started with this? Please and thank you. 😊

PR #958, opened 2023-03-16
governance/core: make naming convention configurable

simongottschlag:
This PR introduces the usage of the provider [aztfmod/azurecaf](https://registry.terraform.io/providers/aztfmod/azurecaf/latest/docs) which makes naming of Azure resources more configurable.

The idea is that we make prefixes and suffixes of resources names configurable using [azurecaf_name](https://registry.terraform.io/providers/aztfmod/azurecaf/latest/docs/resources/azurecaf_name) through variables (**NOTE**: we are using the `data` source, not the `resource`). If the variables are not configured, our defaults will be applied. Since we can't nest variables (using one variable as the default for another) we also have to use some `local` black magic to get it working.

The main idea is that someone should be able to configure everything very granularly and by doing it the way I've done it we enable someone to configure **ONLY** suffixes or prefixes for a specific resource type. As an example, someone would be able to disable suffixes for `azure_key_vault` and that would remove the unique suffix from only the key vault resources but nothing else. 

Adding the following to the `governance-regional` module would disable suffixes only for `azurerm_key_vault`:

```terraform
  resource_name_overrides = {
    azurerm_key_vault = {
      suffixes = []
    }
  }
```

This has the negative side effect that if someone wants to change something on **EVERYTHING** (like adding something after `environment`) it would need to be configured on each of the resources. This is something we could solve with even more logic, but not sure if it's worth the additional complexity.

My opinion is that the readability of most things actually improve using this change, but the complexity is a bit worse. We don't need the huge string interpolations for Azure AD groups and applications anymore which makes them a bit more sane to look at.

I've been testing it out in the lab and seems to be working OK, but see this as an example to discuss it's merits and if we want to do this for governance and core.

Related to #501


simongottschlag:
I've made a change based on feedback from @phillebaba, changing from `resource` to `data` which makes this a non-breaking change. I've also moved all the `locals` black magic to a separate module.

simongottschlag:
I've made the changes I could find for the `core` module as well. Please review and discuss.

phillebaba (approved):
From my perspective this is both good and bad. Replacing the string templates with a resource is great. I am however not the biggest fan of the way variables are customized. Then again I can't think of a better solution for this.

I have approved as the pros outweigh the cons.

PR #959, opened 2023-03-27
Update node-local-dns and move to using registry.k8s.io

phillebaba:
We need to migrate to the new Kubernetes image registry. Updated the image version while I am at it.


CalleB3 (approved):
LGTM

PR #968, opened 2023-04-06
Move Datadog to install with Flux

phillebaba:


modules/kubernetes/datadog/main.tf
```diff
@@ -14,15 +14,149 @@ terraform {
...
+  namespace: flux-system
+spec:
+  interval: 5m
+  sourceRef:
+    kind: GitRepository
+    name: flux-system
+  path: ./platform/we-dev-aks1/datadog
+  prune: true
+  validation: client
+ EOF
```
simongottschlag, line 39: I don't like doing these as strings. Either HCL to YAML or using templates. This feels like things like number of spaces will cause issues.

modules/kubernetes/datadog/templates/platform.yaml.tpl
```diff
@@ -1,11 +1,55 @@
...
+spec:
+  chart:
+    spec:
+      chart: datadog-operator
+      sourceRef:
+        kind: HelmRepository
+        name: datadog
+      version: 0.8.0
+  values:
+    crds: CreateReplace
```
phillebaba, line 33: Remove line

modules/kubernetes/datadog/variables.tf
```diff
@@ -32,5 +37,5 @@ variable "namespace_include" {
 variable "apm_ignore_resources" {
   description = "The resources that shall be excluded from APM"
   type        = list(string)
-  default     = []
+  default     = [""]
```
phillebaba, line 40: Why the empty string in list?

CalleB3:
Moved to #972

PR #972, opened 2023-04-13
Update Datadog to install with flux

CalleB3:
This alos includes a bump of the datadog-operator to v0.9.2, this to ease the upgrade to v1.0.0 after this has been merged.
And bumping git-provider to 0.0.2


modules/kubernetes/datadog/templates/platform.yaml.tpl
```diff
@@ -70,4 +113,4 @@ spec:
       enabled: true
     logCollection:
       enabled: true
-      logsConfigContainerCollectAll: true
+      logsConfigContainerCollectAll: true
```
phillebaba, line 116: Add new line

modules/kubernetes/datadog/templates/datadog-operator.yaml.tpl
```diff
@@ -0,0 +1,42 @@
...
+  chart:
+    spec:
+      chart: datadog-operator
+      sourceRef:
+        kind: HelmRepository
+        name: datadog
+      version: 0.8.0
+  values:
+    appKey: ${app_key}
+    apiKey: ${api_key}
```
phillebaba, line 34: This is going to be an issue. We need to have the operator read this from a secret, as we cannot store it in Git.

phillebaba: A CSI Secret mount needs to be created which results in the following Secrets.
https://github.com/DataDog/helm-charts/blob/d9204cfb44e26cf8ab3911472930ec51a24d02d5/charts/datadog-operator/templates/deployment.yaml#L65-L78

phillebaba (approved):

PR #973, opened 2023-04-13
Switch to Standard sku_tier due to deprecations in the AzureAPI

### LINKED ISSUE #967: SKU Tier and SKU Name changing for AKS
To future-proof and with Azure changing the API Version for AKS, they are changing the SKUName and SKUTier. We need to make corresponding change in the modules.

image

Currently waiting for AzureRM to start using the new API version so we can perform the changes needed. See:
https://github.com/hashicorp/terraform-provider-azurerm/issues/20887

CalleB3:
Azure API Version 2023-02-01 and later now require sku_tier to be set to Standard instead of Paid.

Fixes #967


phillebaba (approved):
LGTM

PR #985, opened 2023-05-04
Manage Flux notification provider

phillebaba:
It seems more reasonable for the provider to be automatically created in the tenants namespace. The configuration is very similar to the GitRepository that is created so this removes some burden on end users.


CalleB3 (approved):
LGTM

PR #995, opened 2023-06-15
Fix Kubernetes version validation

phillebaba:
EKS version validation was not updated. This change fixes this and locks support to only Kubernetes 1.24 and 1.25.


CalleB3 (approved):
LGTM

PR #1020, opened 2023-08-03
azure/governance: Fixes #1015: Added ability to configure azure disk encryption on key vault

### LINKED ISSUE #1015: Add option to set enabled_for_disk_encryption
Sometimes when running the governance module the key vault want to remove its configuration for allowing ADE to access the key vault for BEK/KEK keys

```ARM
  # module.governance_regional.azurerm_key_vault.delegate_kv["core"] will be updated in-place
  ~ resource "azurerm_key_vault" "delegate_kv" {
      - enabled_for_disk_encryption     = true -> null
        id                              = "/subscriptions/<sub-id>/resourceGroups/rg-prod-we-core/providers/Microsoft.KeyVault/vaults/<kv-name>"
        name                            = "<kv-name>"
        # (11 unchanged attributes hidden)

        # (1 unchanged block hidden)
    }
```

It wants to remove this setting: 

image

Can we have the posibility to manually set this setting to `enabled_for_disk_encryption = true` to prevent this from happening? It could be optional with default set to false as not all key vaults should have this activated. 

Thanks in advance!

carlzxc71:
This PR is in reference to this open issue: https://github.com/XenitAB/terraform-modules/issues/1015

**Background**

If Azure Disk Encryption (ADE) has been enabled on a key vault, running the governance module will attempt to disable this setting. Whilst you could run terraform and manually enable it again in the portal it is not a great solution. 

**Fix**

I have adjusted the module slightly and have followed all the steps in the contribution guide as well as tested my branch against the environment where I have had the issue with success. I can confirm that this allowed me to configure the ADE settings to my liking without affecting other settings or other key vaults.


carlzxc71:
Changes required for consumer: 

Update the variable `resource_group_configs` in `variables.tf`

```HCL
variable "resource_group_configs" {
  description = "Resource group configuration"
  type = list(
    object({
      common_name                        = string
      delegate_aks                       = bool # Delegate aks permissions
      delegate_key_vault                 = bool # Delegate KeyVault creation
      delegate_service_endpoint          = bool # Delegate Service Endpoint permissions
      delegate_service_principal         = bool # Delegate Service Principal
      lock_resource_group                = bool # Adds management_lock (CanNotDelete) to the resource group
      disable_unique_suffix              = bool # Disable unique_suffix on resource names
      key_vault_purge_protection_enabled = optional(bool, false)
      enabled_for_disk_encryption        = optional(bool, false)
      tags                               = map(string)
    })
  )
}
```

Update `common.tfvars`

```HCL
resource_group_configs = [
  {
    common_name                        = "core",
    delegate_aks                       = false,
    delegate_key_vault                 = true,
    delegate_service_endpoint          = false,
    delegate_service_principal         = false,
    lock_resource_group                = false,
    disable_unique_suffix              = false,
    key_vault_purge_protection_enabled = false,
    enabled_for_disk_encryption        = true,
    tags = {
      "description" = "Core infrastructure"
    }
  },
]
```

phillebaba:
@carlzxc71 is there any cost associated with enabling `enabled_for_disk_encryption`. If the answer is no I would suggest just always having it enabled.

phillebaba:
I want to keep the configuration as simple as possible, it should not be an configuration if there is no need to disable it. Change the PR to just hard code this parameter and I will approve it.

carlzxc71:
@phillebaba - Updated.

phillebaba (approved):
LGTM

phillebaba (approved):

PR #1027, opened 2023-08-30
Add purge task to remove old images from ACR

phillebaba:
This task will purge images that are older than 365 days, with the exception that it will keep the 5 newest versions. This will  remove the risk of removing images that have not been updated for a year but are still being used.


CalleB3 (approved):
LGTM

PR #1048, opened 2023-12-12
Add azad kube proxy password to core key vault

landerss1:
With the azad kube proxy password in the key vault, the kubernetes/aks-core module can be separated from the rest of aks, significantly reducing blast radius and the size of the terraform state, thereby simplifying updates of the cluster.


landerss1:
Accidentally included changes intended for separate PR.

PR #1049, opened 2023-12-12
Add azad kube proxy password in core key vault

landerss1:
Add azad kube proxy password to the core key vault. This will enable the separation of the kubernetes/aks_core module from the rest of aks, thereby minimizing blast radius and simplifying maintenance of aks.


CalleB3 (approved):
LGTM

PR #1050, opened 2023-12-12
Migrate diagnostic settings

landerss1:
[Migrate diagnostic settings storage retention to Azure Storage lifecycle management - Azure Monitor](https://learn.microsoft.com/en-us/azure/azure-monitor/essentials/migrate-to-azure-storage-lifecycle-policy?tabs=portal).


CalleB3 (approved):
LGTM

PR #1054, opened 2024-02-01
Migrate to opentofu

landerss1:
Fix error with object constructor map keys must be unique to facility upgrade to 1.6.0 and migration to OpenTofu.


CalleB3 (approved):
LGTM

PR #1055, opened 2024-02-02
Add functionality for multiple flux-tenants in one environment

CalleB3:


modules/kubernetes/fluxcd-v2-azdo/templates/tenant.yaml
```diff
@@ -52,6 +53,24 @@ spec:
...
+---
+apiVersion: kustomize.toolkit.fluxcd.io/v1beta2
+kind: Kustomization
+metadata:
+  name: ${name}
+  namespace: ${name}
+spec:
+  serviceAccountName: flux
+  interval: 5m
+  path: ./tenant/${environment}/${name}
```
phillebaba, line 67: I think that you could just put the if statement around the path instead of duplicating the Kustomization.

CalleB3: Will change that

phillebaba (approved):

PR #1066, opened 2024-03-28
Enable workload identity service account in tenant namespaces

phillebaba:
This change will create a new user assigned identity with a configured federated identity to enable tenant workload identities. The reason a new identity is created is because it now has to be set per cluster. This change is not breaking as it will keep the existing aad pod identity solution until not needed anymore.

### DIFF
--- a/modules/azure/aks/workload-identity.tf
+++ b/modules/azure/aks/workload-identity.tf
@@ -1,0 +1,45 @@
+resource "azurerm_user_assigned_identity" "tenant" {
+  for_each = {
+    for ns in var.namespaces :
+    ns.name => ns
+  }
+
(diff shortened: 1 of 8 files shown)


modules/azure/aks/workload-identity.tf
```diff
@@ -0,0 +1,52 @@
...
+    for ns in var.namespaces :
+    ns.name => ns
+  }
+  resource_group_name = data.azurerm_resource_group.this.name
+  location            = data.azurerm_resource_group.this.location
+
+  name = "uai-${var.environment}-${var.location_short}-${var.name}${local.aks_name_suffix}-${each.key}"
+}
+
+resource "azurerm_role_assignment" "tenant" {
```
phillebaba, line 12: Dont think this is actually needed anymore.

landerss1 (approved):
LGTM

PR #1067, opened 2024-04-03
Update cert-manager to use workload identities for authentication

phillebaba:
This adds a new user assigned identity as it was using the same resources as external-dns before. This change also updates the cert-manager version as the older version does not support the new auth method.


modules/azure/aks/variables.tf
```diff
@@ -204,3 +204,13 @@ variable "log_eventhub_authorization_rule_id" {
   description = "The authoritzation rule id for event hub"
   type        = string
 }
+
+variable "dns_zones" {
```
CalleB3, line 208: might wanna call this `dns_zone` to keep the same naming in every module?

phillebaba: The name for all other variables is wrong. Was planning on updating the other modules. This is a list of zones not a singe zone.

CalleB3: True, that makes more sense

CalleB3 (approved):
LGTM

PR #1071, opened 2024-04-05
Add support for Microsoft Defender for containers

landerss1:
This PR adds support for Microsoft Defender for containers as an alternative to Trivy. It also adds the option of sending Kubernetes audit logs to an analytics workspace where it can be consumed by Sentinel.


modules/azure/aks/defender.tf
```diff
@@ -0,0 +1,117 @@
+resource "azurerm_security_center_auto_provisioning" "this" {
+  auto_provision = "Off"
+}
+
+resource "azurerm_security_center_subscription_pricing" "containers" {
+  count         = var.defender_enabled ? 1 : 0
```
phillebaba, line 6: Use for each instead of counts.

modules/azure/aks/variables.tf
```diff
@@ -205,6 +205,47 @@ variable "log_eventhub_authorization_rule_id" {
...
+  default     = false
+}
+
+variable "audit_config" {
+  description = "Kubernetes audit log configuration"
+  type = object({
+    destination_type = optional(string, "StorageAccount")
+    analytics_workspace = optional(object({
+      sku_name       = optional(string, "PerGB2018")
+      daily_quota_gb = optional(number, -1)
```
phillebaba, line 220: Does -1 mean infinite?

landerss1: Yes. Do we have a better number to use?

modules/azure/aks/defender.tf
landerss1: What is the advantage?

modules/azure/aks/variables.tf
phillebaba: I just mean does it also mean infinite for the API.

modules/azure/aks/defender.tf
phillebaba: This is one of the few cases where count could be accepted, but we have been avoiding it for a long time now. So its just a design choice.

modules/azure/aks/variables.tf
landerss1: It means "The workspace daily quota for ingestion in GB", so no.

modules/azure/aks/defender.tf
landerss1: OK, I think it looks weird to use a for-each on something that is on or off, but sure, I can fix that.

phillebaba (dismissed):
LGTM

phillebaba (approved):

PR #1073, opened 2024-04-09
Remove provider aws

landerss1:
This PR removes support for the AWS cloud provider.


modules/kubernetes/ingress-nginx/templates/values.yaml.tpl
```diff
@@ -25,23 +25,12 @@ controller:
...
-  # Optionally change this to ClusterFirstWithHostNet in case you have 'hostNetwork: true'.
-  # By default, while using host network, name resolution uses the host's DNS. If you wish nginx-controller
-  # to keep resolving names inside the k8s network, use ClusterFirstWithHostNet.
-  dnsPolicy: ClusterFirstWithHostNet
-  hostNetwork: true
-  %{~ endif ~}
-
   service:
     externalTrafficPolicy: Local
-    %{~ if provider == "aws" || internal_load_balancer || external_dns_hostname != "" ~}
```
phillebaba, line 38: We can keep this but I am not sure if it is beeing used in Azure.

modules/kubernetes/cluster-autoscaler/templates/values.yaml.tpl
```diff
@@ -1,4 +1,4 @@
-cloudProvider: ${provider}
```
phillebaba, line 1: Cluster autoscaler module should be completely removed as it is not used in Azure. AKS runs this for you.

phillebaba (commented):
End of an era. Things looks good and we are bound to have missed

CalleB3 (approved):

modules/kubernetes/ingress-nginx/templates/values.yaml.tpl
landerss1: Just removed the aws condition, so should keep working.

modules/kubernetes/cluster-autoscaler/templates/values.yaml.tpl
```diff
@@ -1,4 +1,4 @@
...
+cloudProvider: "azure"
```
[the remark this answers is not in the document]
landerss1: Done.

modules/kubernetes/cluster-autoscaler/templates/values.yaml.tpl
landerss1: Done, History.

phillebaba (approved):
LGTM

PR #1077, opened 2024-04-09
Minimum retention days for analytics workspace

landerss1:
Minimum retention days for analytics workspace is 30 for sku PerGB2018


CalleB3 (approved):
LGTM

PR #1078, opened 2024-04-10
Remove delegate resource group from namespaces

phillebaba:
I cannot find any use where delegate_resource_group is false. This just adds complexity to the namespace configuration so we should just completely remove it instead.


landerss1 (approved):
Could have changed the type of the `namespaces `variable to `list(string)` instead, but let's keep the object type for now in case we need to add som other attributes in the future. Also less changes required for the tenant clusters.

PR #1080, opened 2024-04-12
Azuread deprecations introduced in provider v 2.44

landerss1:
Handle deprecations introduced in version 2.44 of the azuread provider.


CalleB3 (approved):
LGTM

PR #1084, opened 2024-04-16
Make it possible to use CoreDNS as the last route in node-local-dns

CalleB3:


landerss1 (approved):

modules/kubernetes/aks-core/variables.tf
```diff
@@ -647,3 +647,9 @@ variable "defender_enabled" {
   type        = bool
   default     = false
 }
+
+variable "use_coredns" {
```
phillebaba, line 651: I suggest a different name for this. Maybe coredns_upstream or something similar instead.

CalleB3: Indeed, that is better.

phillebaba (approved):

landerss1 (approved):

PR #1092, opened 2024-04-24
refactor: create new Fluxcd module

phillebaba:


phillebaba:
The main challenge with getting this to work is to figure out how to get two different credentials running with the provider. One set of credentials that the provider uses to authenticate, and another one which will use git-auth-proxy to sync. This has to be acheived through patching the existing configuration before it is applied.

landerss1:
Implemented in [1206](https://github.com/XenitAB/terraform-modules/pull/1206) and related fixes:

- 1207-1209
- 1211-1214

Closing this.

PR #1105, opened 2024-05-16
Fix: Velero storage account name format

CalleB3:
This fixes issue with `-` in the name for velero storage account, also makes the name of the SA and SA container configurable if one wants/needs to do that.


landerss1 (approved):
LGTM

landerss1 (approved):

PR #1110, opened 2024-05-21
Fix: add a private ingressclass instead of replacing the existing

CalleB3:
This PR keeps the `nginx` ingress class when enabling private ingress.
Instead of replacing `nginx` with `nginx-public`

Making the workflow to enable it alot easier if you already use the `nginx` class.

Also making the logic in azad-kube-proxy somewhat simpler.


landerss1 (dismissed):
LGTM

landerss1 (approved):

PR #1140, opened 2024-06-19
refactor: update of the default log retention time for azure to 30 days

yabracadabra:
refractor: change of default value for log retention from 1 year for all environments to 30 days. This regards all environments to have the dev and qa use 30 as default and prod to still keep using 365 days by using the variable according to ISO27001.


CalleB3 (approved):
LGTM

PR #1151, opened 2024-08-05
feat: grafana-alloy module for XKS deployment

optocoupler:


landerss1 (changes requested):
There is a kubernetes_namespace resource with the same name as the namespace declared in the template. Having the kubernetes_namespace resource is the way we used to do it, and we should probably revert to that method as having Flux manage the namespace has created issues. So, I suggest you remove the namespace resource from the template, and declare a dependency to the tofu resource.

PR #1158, opened 2024-08-16
feat: prevent destruction of tenant namespaces

landerss1:
This PR adds  a lifecycle management tag to tenant namespace resources that will prevent the namespace from being destroyed. From now on, destorying tenant namespaces can only be made outside of Terraform/Open Tofu.


optocoupler (approved):
LGTM

PR #1167, opened 2024-09-04
feat: add initial support for automation alerts

landerss1:
This PR adds rudimentary support for alerting when a runbook job fails. Initially, only email receivers are supported.


CalleB3 (approved):
Lgtm

PR #1168, opened 2024-09-09
refactor: explicitly set AKS SKU tier and node count

landerss1:
This PR changes the way we specify the AKS SKU tier and number of nodes in the default node pool. Instead of inferring this from an attribute called `production_grade,` we explicitly set this in the config to allow more flexibility.


optocoupler (approved):
LGTM

PR #1191, opened 2024-10-02
chore: Update Ingress-nginx, Cert-manager and external-dns

CalleB3:
Tested and all works fine in unbox-sand cluster.


landerss1 (approved):
LGTM

PR #1196, opened 2024-10-04
feat: Move azad-kube-proxy to aks-core and use SecretProviderClass

CalleB3:
Sort of breaking; needs to run AKS-regional to remove everything before reapplying aks-core
OR: remove from state and import. but the disruption should be OK

This PR moved all azad-kube-proxy related config to aks-core module. Keeping it all in one place.

Also adds Workload Identity and proper secrets management.


landerss1 (approved):
LGTM

optocoupler (approved):
LGTM

PR #1203, opened 2024-10-10
chore!: bump flux provider to v1.4.0

landerss1:
This PR introduces **breaking changes** with the upgrade of the FluxCD provider to v1.4.0. Also, it replaces the two previous modules, `fluxcd-v2-azdo` and `fluxcd-v2-github` with a brand new `fluxcd` module that handles both Azure DevOps and GitHub. To migrate an existing cluster, one should follow the example below (Azure DevOps) for the **aks-core** module:

1. `terraform destroy module.aks1_core.module.fluxcd_v2_azure_devops["fluxcd-v2"].helm_release.git_auth_proxy`
2. Reconfigure your Terraform code according to the changed variables related to Flux
3. `terraform import module.aks1_core.module.fluxcd[\"fluxcd\"].kubernetes_namespace.flux_system flux-system`
4. `terraform state rm module.aks1_core.module.fluxcd_v2_azure_devops`
5. `terraform plan` & `terraform apply`

As for p.2 above, check the variables defined in [modules/kubernetes/aks-core/variables.tf](https://github.com/XenitAB/terraform-modules/compare/main...flux-upgrade#diff-19c9d4015e24bda6804a37561ebfa4ad996e2327c0308cdd80fb2c1b5ea60f96):

- variable `namespaces` has changed
- variable `fluxcd_v2_enabled` has changed name to `fluxcd_enabled`
- variable `fluxcd_v2_config` has changed name to `fluxcd_config` and has changed


modules/kubernetes/fluxcd/main.tf
```diff
@@ -0,0 +1,100 @@
...
+  values = [templatefile("${path.module}/templates/git-auth-proxy-values.yaml.tpl", {
+    git_provider = var.git_provider
+    bootstrap    = var.bootstrap
+    tenants      = var.namespaces
+  })]
+}
+
+resource "kubernetes_namespace" "flux_system" {
+  metadata {
+    name = "flux-system"
```
optocoupler, line 58: We should add the labels for the namespace like the others.

optocoupler (changes requested):
Just a small addition to add the label for the flux-system namespace and the rest looks OK.

modules/kubernetes/fluxcd/main.tf
landerss1: Nope. Flux generates a Namespace manifest as part of the bootstrap, which contains the labels.

optocoupler (approved):
LGTM

PR #1226, opened 2024-10-23
feat: add support for Gateway API

landerss1:
This PR adds initial support for the Gateway API. It:

- Deploys CRDs from the Gateway API standard or experimental [channel](https://github.com/kubernetes-sigs/gateway-api/tree/main/config/crd)
- Configures external-dns to support Gateway API
- Configures cert-manager to support Gateway API

Gateway providers that implement the Gateway API, such as Contour or Nginx Gateway Fabric, are expected to be deployed separately and depend on this module to be deployed first.


optocoupler (approved):
LGTM

PR #1236, opened 2024-11-04
feat: add envoy gateway

optocoupler:


modules/kubernetes/aks-core/variables.tf
```diff
@@ -242,6 +242,11 @@ variable "azure_policy_config" {
     ]
   }
 }
+variable "envoy_tls_policy_enabled" {
+  description = "An option to remove the gatekeeper mutation for tls settings"
+  type        = bool
+  default     = true
```
landerss1, line 248: Can we pls have this set to false instead?

modules/kubernetes/aks-core/variables.tf
```diff
@@ -834,6 +839,23 @@ variable "karpenter_config" {
   }
 }
 
+
+variable "envoy_gateway_enabled" {
+  description = "Should we deploy envoy-gateway"
+  type        = bool
+  default     = false
+}
```
landerss1, line 847: Can we pls have a space between variable declarations?

modules/kubernetes/azure-policy/locals.tf
```diff
@@ -17,6 +17,11 @@ locals {
     templatefile("${path.module}/templates/flux-require-service-account.yaml.tpl", {
     })
   )
+  envoy_gateway_require_tls = base64encode(
```
landerss1, line 20: Can we pls have a blank line before and after variable declarations?

modules/kubernetes/azure-policy/main.tf
```diff
@@ -279,6 +279,81 @@ resource "azurerm_policy_definition" "flux_require_service_account" {
...
+        "type": "Array",
+        "metadata": {
+          "displayName": "Namespace inclusions",
+          "description": "List of Kubernetes namespaces to only include in policy evaluation. An empty list means the policy is applied to all resources in all namespaces."
+        },
+        "defaultValue": []
+      }
+    }
+    PARAMETERS
+}
```
landerss1, line 356: Can we pls have a blank line after resource declaration?

landerss1 (changes requested):
Looks good! Just minor changes please. See comments.

modules/kubernetes/aks-core/variables.tf
optocoupler: I don't think thats the best; I want users to make a conscious decision to disable the policy, i.e. have potentially insecure settings deployed in their cluster.

landerss1: Yeah, but for those customers that use azure policy, and not envoy, there will be mutations created anyway. In order to avoid that, we would have to add the variable in every architecture that doesn't use envoy. I prefer making sure that introducing new functionality and new variables doesn't require that we add those variables to existing infrastructures. Ideally, the infrastructures should be untouched by new functionality.

landerss1 (approved):

landerss1 (approved):
LGTM

PR #1239, opened 2024-11-08
feat: add support for popeye k8s linter

landerss1:
This PR adds support for the [popeye](https://github.com/derailed/popeye) k8s live cluster linter.


optocoupler (approved):
Good to go!

PR #1252, opened 2024-11-14
refactor!: create required AD-groups in governance module

landerss1:
This PR adds support for creating required AD groups as part of the governance module, instead of having to create them manually in advance. The following AD-groups are now created automatically:

- az-sub-<subscription_name>-all-owner
- az-sub-<subscription_name>-all-contributor
- az-sub-<subscription_name>-all-reader

Existing AD-groups must be imported using

```
terraform import module.governance_global.azuread_group.all_owner["delegate_sub_groups"] 00000000-0000-0000-0000-000000000000
terraform import module.governance_global.azuread_group.all_contributor["delegate_sub_groups"] 00000000-0000-0000-0000-000000000000
terraform import module.governance_global.azuread_group.all_reader["delegate_sub_groups"] 00000000-0000-0000-0000-000000000000
```


landerss1:
This PR overlooked the fact that these AD-groups are truly global across clusters, so they would be overwritten by other clusters usage of the governance modules. Closing.

PR #1266, opened 2025-01-20
fix: Upgrade Velero and plugin to support Workload Identity

CalleB3:
Since migration of Velero was done to use Workload Identity, it hasnt worked since it wasnt supported in the old version we were running.
only upgraded it to the point were Workload Identity was introduced.

This change upgrades the chart and azure-plugin to solve that.


optocoupler (approved):
lgtm

PR #1278, opened 2025-01-30
fix: hardcode kubelet_disk_type to OS

CalleB3:


landerss1:
The attribute is already optional with OS as default. No need to hardcode the value. Closing this PR.

PR #1297, opened 2025-02-19
fix: aks validation should allow no nodepools elements

landerss1:
This PR adjusts the aks validation to allow the creation of a cluster without any nodepools, except the system one.


optocoupler (approved):
lgtm

PR #1303, opened 2025-02-21
feat: add support for removing spot instances taints in clusters that are automated to start/stop

landerss1:
This PR adds a temporary hack to remove spot instances taints in clusters that are stopped and started using the xks automation feature.


optocoupler (approved):
LGTM

PR #1318, opened 2025-03-27
feat: add delegations block to subnet

linus-lysell:
Add delegation block in subnet resource as optional to be able to configuring private networking for GitHub-hosted runners in your enterprise.


landerss1 (approved):
LGTM

PR #1319, opened 2025-03-27
feat: add support for nsg rule destination/source_address_prefixes

linus-lysell:
Added support for destination_address_prefixes and source_address_prefixes in nsg rule configuration.


landerss1 (approved):
LGTM

PR #1337, opened 2025-05-15
fix: adding an additional ingress block in the default-deny

optocoupler:
…olicy to also allow traffic to the namespace from grafana collectors


yabracadabra:
#1337

PR #1357, opened 2025-06-18
feat: add tenant support for ArgoCD

landerss1:
This PR adds support for deployment of ArgoCD to tenant clusters. The deployment can be in each cluster, or as a hub-spoke architecture.


optocoupler (approved):
lgtm

optocoupler (approved):
now it looks even better

PR #1393, opened 2025-10-23
fix: namespace exclusions in Azure Policy

CalleB3:


CalleB3:
Not sure if we wanna exclude namespaces from all constraints if used, or only specific ones.

PR #1463, opened 2026-04-28
fix: change Falco driver kind from ebpf to auto

CalleB3:
`kind: modern_ebpf` seem to be the preferred way to go, `auto` will solve this for us.


yabracadabra (approved):
LGTM

